PR #2, opened 2019-02-03
Hoover

mgax:
Run hoover in a liquid node under nomad.


salevajo:
Checked this PR #2 on a new installation successfully.

PR #6, opened 2019-02-27
added nextcloud.nomad

MattesR:
Here's the part of my docker-compose file, which should be recreated as a nomad file

https://gist.github.com/MattesR/0165f1122090c210d22630ee3c81a6af


nomad/nextcloud.nomad
```diff
@@ -0,0 +1,261 @@
...
+  type = "service"
+  
+  group "nc" {
+    task "nextcloud" {
+      driver = "docker"
+      config {
+        image = "nextcloud:15"
+        volumes = [
+          "/opt/hoover/volumes/nextcloud:/var/www/html",
+          "/opt/hoover/collections/ncsync:/var/www/html/data/ncsync/files",
```
mgax: In the other nomad jobs, volumes are stored under `/var/local/liquid/volumes/{job}/{group-or-task}`, we should keep a consistent place. Maybe we even make it configurable someday.

nomad/nextcloud.nomad
```diff
@@ -0,0 +1,261 @@
...
+          port "http" {}
+        }
+      }
+      service {
+        name = "nextcloud"
+        port = "http"
+      }
+      }
+
+    task "pg" {
```
mgax: I generally prefer to have dependencies at the top, but it makes sense to have them lower, out of the way, so it's easy to see the main service. What do you think?

nomad/nextcloud.nomad
```diff
@@ -0,0 +1,261 @@
...
+      resources {
+        network {
+          port "http" {}
+        }
+      }
+      service {
+        name = "nextcloud"
+        port = "http"
+      }
+      }
```
mgax: Argh indentation :)

nomad/nextcloud.nomad
```diff
@@ -0,0 +1,261 @@
...
+        labels {
+          liquid_task = "nextcloud-pg"
+        }
+        port_map {
+          pg = 5432
+        }
+      }
+      env {
+        POSTGRES_USER = "postgres"
+        POSTGRES_PASSWORD = "secret"
```
mgax: Let's add `TODO`s so we remember to change the secrets. (we can store them in [vault](https://www.vaultproject.io/))

nomad/nextcloud.nomad
```diff
@@ -0,0 +1,261 @@
...
+        destination = "local/snoop.env"
+        env = true
+      }
+      resources {
+        memory = 512
+      }
+    }
+  }
+  group "api" {
+    task "snoop" {
```
mgax: Do you really need a `snoop` task in the nextcould job? I think the snoop should be separate, and configured to pull data from nextcloud via nomad *services*.

nomad/nextcloud.nomad
MattesR: Configurable would be best. I just hacked it together like this because I didn't want to move all my files ^^
I didn't run it on any nomad cluster since I couldn't get Nomad to run in my VMs yet. I took your nomad files for the testdata collection and for hoover and just made a resembling file.

nomad/nextcloud.nomad
MattesR: I did it like this so that it is the whole package in one file.  I also set them up with the same script. We could change that but I saw them as intertwined so much that bringing it all up includes snoop and nextcloud.

nomad/nextcloud.nomad
MattesR: Is there a Nomad equivalent to `depends_on:`?  
Actually I can see the benefit from having them at the top too.

nomad/nextcloud.nomad
mgax: Yes, but we already have a nomad config for snoop, it would be problematic to maintain two in parallel. I'd rather have the one be configurable so it can pull data from nextcloud.

nomad/nextcloud.nomad
mgax: There is no direct equivalent. But you can link to service dependencies through consul's service registry, like we already have between hoover and snoop, and in other places.

nomad/nextcloud.nomad
mgax: What went wrong? We could do a debugging session after the call.

nomad/nextcloud.nomad
MattesR: You're right about that.  
Come to think of it, I'll add an issue to the board about collection creation with the nomad setup.

liquid.py
```diff
@@ -204,6 +204,14 @@ def shell(name, *args):
     docker_exec_cmd = ['docker', 'exec', '-it', id] + list(args or ['bash'])
     run_fg(docker_exec_cmd, shell=False)
 
+def usershell(user, name, *args):
+    """
+    Open a shell in a docker container tagged with liquid_task=`name` as 'user'
```
mgax: Ok, why don't we add a flag to the `shell` command? It does receive all arguments in `args` and `argparse` is good at parsing them. Something like:

```python
import argparse
parser = argparse.ArgumentParser()
parser.add_argument('-u', '--username')
parser.add_argument('cmd', nargs=argparse.REMAINDER)
options = parser.parse_args(args)
```

This creates a potential to capture `-u` flags from the command, but a user can work around that:

```shell
./liquid.py shell -- do_something.sh -u other  # `-u` is passed on to `do_something.sh`
```

collection-ncsync.nomad
```diff
@@ -0,0 +1,201 @@
+job "collection-ncsync" {
```
mgax: Once https://github.com/liquidinvestigations/node/pull/17 is merged, we should use collection templates for the synced collection.

Readme.md
```diff
@@ -71,6 +71,33 @@ collection:
 ./liquid.py shell hoover-search ./manage.py addcollection testdata --index testdata http://$(./liquid.py nomad_address):8765/testdata/collection/json --public
 ```
 
+### Nextcloud
+
+First, make directories:
+```shell
+./makedirs
```
mgax: `.sh`

nc_setup/liquid/core/templates/layout.user.php
```diff
@@ -0,0 +1,130 @@
+<!DOCTYPE html>
```
mgax: Do we need to clone all this HTML here? How much different is it from the original `layout.user.php`? My concern is that the upstream version will evolve and it will be difficult to update our copy.

nc_setup/liquid/core/templates/login.php
```diff
@@ -0,0 +1,44 @@
+<?php /** @var $l \OCP\IL10N */ ?>
```
mgax: Same for this template.

nextcloud_setup.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
```
gabriel-v:
```
set -e
set -x
```

makedirs.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
```
gabriel-v:
```
set -e
set -x
```

Readme.md
```diff
@@ -148,6 +148,26 @@ collection:
 ./liquid initcollection testdata
 ```
 
+### Nextcloud
+
+Tell liquid we want to run nextcloud and nc_sync jobs in `liquid.ini`:
+
+```ini
+[job:nextcloud]
+template = templates/nextcloud.nomad
```
mgax: No need for this, just run nextcloud by default. This way travis will try to deploy it too.

Readme.md
```diff
@@ -148,6 +148,26 @@ collection:
...
+
+[collection:ncsync]
```
mgax: I think it's also pretty safe to create this collection by default.

templates/nextcloud.nomad
```diff
@@ -0,0 +1,80 @@
...
+      resources {
+        network {
+          port "http" {}
+        }
+      }
+      service {
+        name = "nextcloud"
+        port = "http"
+      }
+      }
```
mgax: Indentation please :)

templates/nextcloud.nomad
```diff
@@ -0,0 +1,80 @@
...
+
+    task "pg" {
```
mgax: Let's put the task "pg" in a separate group called "db", this way they will be allocated separately on the cluster.

templates/nextcloud.nomad
```diff
@@ -0,0 +1,80 @@
...
+      driver = "docker"
+      config {
+        image = "postgres:9.6"
+        volumes = [
+          "${liquid_volumes}/nextcloud/nextcloud-pg/data:/var/lib/postgresql/data",
```
mgax: I think `"${liquid_volumes}/nextcloud/pg/data:/var/lib/postgresql/data"` is fine too.

templates/nextcloud.nomad
```diff
@@ -0,0 +1,80 @@
...
+        labels {
+          liquid_task = "nextcloud"
+        }
+      }
+      template {
+        data = <<EOF
+            NEXTCLOUD_POSTGRES_HOST =
+              {{- range service "nextcloud-pg" -}}
+                {{ .Address }}:{{ .Port }}
+              {{- end }}
```
mgax: I've changed this pattern in the other templates because it's dangerous. It allows the env var to be set to weird values. This is somewhat safer:
```
{{- range service "nextcloud-pg" }}
  NEXTCLOUD_POSTGRES_HOST = {{.Address}}:{{.Port}}
{{- end }}
```

Notice how only the open braces have whitespace removal (`{{-`) while the closing braces don't (`}}`) - if we stick to this rule we avoid bugs with collapsed newlines that result in concatenated env vars.

templates/nextcloud.nomad
```diff
@@ -0,0 +1,80 @@
...
+            {{- range service "nextcloud-pg" }}
+              NEXTCLOUD_POSTGRES_HOST = {{.Address}}:{{.Port}}
+            {{- end }}
+            NEXTCLOUD_HOST = nextcloud.{{ key "liquid_domain" }}
+            NEXTCLOUD_ADMIN_USER="admin"
+            NEXTCLOUD_ADMIN_PASSWORD="admin"
+            NEXTCLOUD_POSTGRES_DB="nextcloud"
+            NEXTCLOUD_POSTGRES_USER="postgres"
+            NEXTCLOUD_POSTGRES_PASSWORD="secret"
+            OC_PASS="a_very_good_secret"
```
mgax: What does this value do? If it's supposed to be a random secret like Django's `SECRET_KEY` then we can auto-generate it and store it in Vault - take a look at hoover-search and see how it's done there.

MattesR: Yeah it is, I will look at it and change it.

PR #17, opened 2019-04-04
Dynamic collections

mugurrus:


collection.nomad
```diff
@@ -1,4 +1,4 @@
-job "collection-testdata" {
+job "collection-${collection_name}" {
```
mgax: Could we call it `$name` instead of `${collection_name}`?

liquid.py
```diff
@@ -27,58 +27,63 @@
...
+    if section_name in config and key in config[section_name]:
+        return config[section_name][key]
 
     return default
 
 
 nomad_url = get_config(
     'NOMAD_URL',
-    'cluster:nomad_url',
+    'cluster.nomad_url',
```
mgax: Yes, `.` is a better separator.

liquid.py
```diff
@@ -27,58 +27,63 @@
...
 
-def get_config(env_key, ini_path, default):
-    value = os.environ.get(env_key)
-    if value is not None:
-        return value
+def get_config(env_key, ini_path, default=None):
+    if env_key and os.environ.get(env_key):
+        return os.environ[env_key]
+
+    if ini_path.endswith(':'):
```
mgax: This adds a lot of complexity to `get_config`. It looks like the end goal is to get collections from `liquid.ini`, which can be achieved outside of this function, with another function e.g. like this:

```python
def get_collections():
    for section_name in config:
        if section_name.startswith('collection:'):
            section = config[section_name]
            yield {
                "name": section_name.split(':')[1],
                "workers": section.get('workers', '1'),
            }
```

liquid.py
```diff
@@ -226,14 +231,33 @@ def nomad_address():
...
+    return hcl.replace('__LIQUID_COLLECTIONS__', liquid_collections)
+
+
+def run_collection_job(name, settings):
+    set_collection_defaults(name, settings)
+
+    with open('collection.nomad') as collection_file:
+        template = Template(collection_file.read())
+
+        hcl = template.substitute(settings)
```
mgax: We're doing two passes of substitution, one for `__LIQUID_*__` and another for the collection parameters. Why don't we do a single pass and convert all our template variables to the same format, either `__VALUE__` or `$value`, whichever feels more readable (I'm partial towards `__`, just because it's clearly different from what a Nomad template looks like).

liquid.py
```diff
@@ -242,16 +266,28 @@ def deploy():
...
     consul.set_kv('liquid_debug', liquid_debug)
 
-    for _, path in JOBS:
+    for job, path in JOBS:
+        print(f'Starting {job}...')
         runjob(path)
 
+    collections = get_config(None, 'collection:')
+    for name, settings in collections:
+        print(f'Starting collection-{name}...')
```
mgax: But we have logging. This should be `log.info("Starting collection-%s ...", name)`.

liquid.py
```diff
@@ -242,16 +266,28 @@ def deploy():
...
     """ Stop all the jobs in nomad. """
 
     for job, _ in JOBS:
+        print(f'Stopping {job}...')
         nomad.stop(job)
 
+    collections = get_config(None, 'collection:')
+    for name, _ in collections:
+        print(f'Stopping collection-{name}...')
+        nomad.stop(f'collection-{name}')
```
mgax: Let's unify the loops:

```python
jobs_to_stop = [j for j, _ in JOBS] + [f'collection-{name}' for name, _ in collections]
for job in jobs_to_stop:
    nomad.stop(job)
```

liquid.py
```diff
@@ -242,16 +266,28 @@ def deploy():
...
 
-    for _, path in JOBS:
+    for job, path in JOBS:
+        print(f'Starting {job}...')
         runjob(path)
 
+    collections = get_config(None, 'collection:')
+    for name, settings in collections:
+        print(f'Starting collection-{name}...')
+        run_collection_job(name, settings)
```
mgax: Same as below, let's have a single loop that calls `runjob`. This will be easier if `runjob` accepts the hcl job content instead of its path.

collection.nomad
```diff
@@ -140,32 +140,32 @@ job "collection-testdata" {
     task "snoop" {
       driver = "docker"
       config {
-        image = "liquidinvestigations/hoover-snoop2"
+        image = "liquidinvestigations/hoover-snoop2:liquid-nomad"
```
gabriel-v: remove this

collection.nomad
```diff
@@ -76,41 +76,41 @@ job "collection-testdata" {
...
 
   group "workers" {
-    count = 2
+    count = $workers
 
     task "snoop" {
       driver = "docker"
       config {
-        image = "liquidinvestigations/hoover-snoop2"
+        image = "liquidinvestigations/hoover-snoop2:liquid-nomad"
```
gabriel-v: remove this

collection.nomad
```diff
@@ -34,10 +34,10 @@ job "collection-testdata" {
       config {
         image = "logicalspark/docker-tikaserver"
         port_map {
-          tika = 9998
+          http = 9998
```
mgax: This was changed to `tika` because `http` was wrong, let's keep the change.

PR #18, opened 2019-04-05
Change elasticsearch data directory permissions when container starts

gabriel-v:
After this we should be able to remove all `chown -R 1000:1000` lines from the [installation manual](https://hackmd.io/MeBET7lQSd2mBOv7vlhwWw#).


mgax:
Tested locally, works! :cactus:

PR #26, opened 2019-04-16
add configuration for dev mount points

mugurrus:


liquid.py
```diff
@@ -62,6 +63,18 @@ def __init__(self):
             'false',
         )
 
+        self.mount_hoover_path = self.get(
+            'LIQUID_MOUNT_HOOVER_PATH',
```
gabriel-v: I think `LIQUID_HOOVER_REPOS_PATH` and `LIQUID_LIQUIDINVESTIGATIONS_REPOS_PATH` are more consistent names with the one above.

liquid.py
```diff
@@ -254,20 +267,42 @@ def set_volumes_paths(substitutions={}):
...
-        ('hoover', 'snoop2', '/opt/hoover/snoop'),
-        ('hoover', 'search', '/opt/hoover/search'),
-        ('liquidinvestigations', 'core', '/app'),
-    ]
-    for (org, repo, target_path) in repos:
-        repo_path = os.path.join(pwd, 'repos', org, repo)
-        repo_volume_line = (f'"{repo_path}:{target_path}",\n')
-        key = f'{org}_{repo}_repo'
+    repos_path = os.path.join(os.path.realpath(os.path.dirname(__file__)), 'repos')
+    repos = {
```
gabriel-v: You can simplify this by extracting some variables:

```python3
li_path = config.mount_li_path or os.path.join(repos_path, 'liquidinvestigations')
hoover_path =  ...

repos = {
    ....
        'local': li_path,
    ....
}
```

And then remove the following two `if`s.

mugurrus: I need `abspath(config.mount_li_path)` (from nomad docs: "Use absolute paths to mount arbitrary paths on the host"), so I can't make this simplification: `abspath(None)` raises exception and treating it would be actually more "expensive".

PR #27, opened 2019-04-17
implement init/purge collection commands

mugurrus:


liquid.py
```diff
@@ -237,17 +260,60 @@ def alloc(job, group):
     print(first(running, 'running allocations'))
 
 
-def nomad_address():
+def collection_exists(name):
+    """Returns true if the collection with the given name was declared in the ini file."""
+
+    names = dict(config.collections).keys()
+    return reduce(lambda exists, curr_name: exists or curr_name.lower() == name.lower(), names, False)
```
gabriel-v: next time you have to write something like this, do
```python3
return any(name.lower() == other.lower() for other in names)
```

liquid.py
```diff
@@ -237,17 +260,60 @@ def alloc(job, group):
...
+    except CalledProcessError as e:
+        if not re.search(f'index \[{name}/[^\]]+] already exists', e.output.decode()):
+            log.error(e.output.decode())
+            raise
+        log.info(f'Index {name} already exists.')
+
+    try:
+        snoop_url = first(nomad.agent_members())['Addr'] + f':8765/{name}/collection/json'
+        shell_cmd('hoover-search', './manage.py', 'addcollection', name, '--index',
+                  name.lower(), snoop_url, '--public')
```
gabriel-v: Don't use `.lower()` here

liquid.py
```diff
@@ -214,14 +225,26 @@ def first(items, name_plural='items'):
     return items[0]
 
 
-def shell(name, *args):
+def shell_cmd(name, *args):
```
mgax: Maybe rename this function to `docker_exec`?

liquid.py
```diff
@@ -237,17 +260,60 @@ def alloc(job, group):
...
+
+
+def nomad_address():
+    print(get_nomad_address())
 
 
 def set_collection_defaults(name, settings):
     settings['name'] = name
-    settings.setdefault('workers', '1')
+    settings.setdefault('workers', config.default_workers_no)
```
mgax: I don't think `'1'` is worth extracting as a default.

liquid.py
```diff
@@ -285,9 +352,83 @@ def get_job(hcl_path, substitutions={}):
...
+            collection_name = job['ID'][len('collection-'):]
+            if collection_name not in collections:
+                log.info('Stopping %s...', job['ID'])
+                nomad.stop(job['ID'])
+
+
+def get_collections_to_purge():
+    """Returns a set of collections to purge."""
+
+    ini_collections = set(dict(config.collections))
```
mgax: Maybe `config.collections` should be an [OrderedDict](https://docs.python.org/3/library/collections.html#collections.OrderedDict) at this point.

liquid.py
```diff
@@ -285,9 +352,83 @@ def get_job(hcl_path, substitutions={}):
...
+
+    try:
+        shell_cmd('hoover-search', './manage.py', 'removecollection', name)
+        log.info(f'Collection {name} was removed from hoover search.')
+    except CalledProcessError as e:
+        if not re.search('DoesNotExist', e.output.decode()):
+            raise
+        log.info(f'No index was found in hoover search for collection {name}.')
+
+    collection_volume = os.path.join(config.liquid_volumes, 'collections', name)
```
mgax: [pathlib](https://docs.python.org/3/library/pathlib.html) please :)

liquid.py
```diff
@@ -285,9 +352,83 @@ def get_job(hcl_path, substitutions={}):
...
+    longer declared in the ini file.
+    """
+
+    to_purge = get_collections_to_purge()
+    if not to_purge:
+        print('No collections need purging.')
+        return
+
+    if to_purge:
+        print('The following collections will be purged:\n-', '\n- '.join(to_purge))
```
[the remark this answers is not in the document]
mgax: Just write a normal loop and print each collection like this: `print('-', collection)`.

liquid.py
```diff
@@ -285,9 +352,83 @@ def get_job(hcl_path, substitutions={}):
...
+    with os.scandir(os.path.join(config.liquid_volumes, 'collections')) as scan:
+        volumes = set([entry.name for entry in scan if entry.is_dir()])
+    return volumes - ini_collections
+
+
+def purge_collection(name):
+    """Purge data for the given collection, remove it from hoover search."""
+
+    try:
+        shell_cmd('hoover-search', './manage.py', 'removecollection', name)
```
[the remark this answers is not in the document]
mugurrus: Collections to be purged are not in the ini file anymore, their snoop containers stopped, so no commands can be run on snoop.

liquid.py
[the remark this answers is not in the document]
mugurrus: Collection name can contain upper case letters, unlike the ES index name. This is a requirement. So when comparing collection names I must compare them as case insensitive.

The problem with allowing the user to write directly to the config file is that he/she can write an invalid name and there's no way to stop him/her. My solution would be to generate an index name that is valid from the collection name, but I'd do this in a separate PR.

liquid.py
mugurrus: I need to, because the collection name might contain upper case letters.

liquid.py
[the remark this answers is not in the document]
mgax: Why is uppercase a requirement? If we're to display user-friendly collection titles, let's make it a separate field, e.g. `title`.

liquid.py
```diff
@@ -214,14 +225,26 @@ def first(items, name_plural='items'):
...
+
+
+def shell(name, *args):
+    """
+    Open a shell in a docker container tagged with liquid_task=`name`
+    """
+    try:
+        shell_cmd(name, *args)
+    except CalledProcessError as e:
+        print(e.output.decode(), file=sys.stderr)
```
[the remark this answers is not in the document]
mugurrus: I needed to print the command output in case of error. Obsolete, as I changed shell to the original functionality.

liquid.py
mugurrus: I'd rather have it in the config

liquid.py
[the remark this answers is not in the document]
mugurrus: I don't see the need for a separate field. The collection name is the title and it may contain upper case letters - this was a requirement from Stephan.

liquid.py
mgax: For sure the workers number should be in the config. But the default can be hardcoded right here, as it was, I think the code was much easier to read.

liquid.py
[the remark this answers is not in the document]
mgax: The collection needs a clean unique machine-readable ID that should avoid any confusion about names of elasticsearch indices and query parameters for the UI. Let's set it explicitly instead of divining it by some undocumented transformation from an arbitrary unicode string.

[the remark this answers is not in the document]
mugurrus: Sorry, I don't understand what you mean by "transformation from an arbitrary unicode string"

AFAIK, there's no confusion regarding the names of ES indices and query parameters for the UI.

For the moment, we should keep the same rules we had on docker-setup for the collection name. Later we can change that if needed.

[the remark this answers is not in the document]
mgax: Ok, I understand your concern, we should be able to migrate existing collections painlessly to this new system. So let's do it the right way, and use the [title](https://github.com/hoover/snoop2/blob/master/snoop/data/views.py#L13) field of snoop's API for the title, as it was intended, but we were too lazy to use a different environment variable.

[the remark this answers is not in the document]
mgax: I made an issue (https://github.com/hoover/snoop2/issues/148) to allow the title to be set via a separate environment variable, so then it becomes easy to define collection titles in `liquid.ini`:

```ini
[collection:testdata]
title = TestData Anÿŧĥĩŋģ ☃️📯
```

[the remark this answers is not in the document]
mugurrus: I suppose you're referring to the collection template. But in what part of that template would you inject the title. Anyway, for this small enhancement I'd rather do a separate PR and not keep adding requirements to this one, in order to keep the PRs small.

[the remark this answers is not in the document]
mgax: Inject the title as env var to snoop api and workers. Ok to make it a separate PR but please drop the `.lower()` for now, so we have clear sane collection IDs.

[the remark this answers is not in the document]
mugurrus: But I can't drop `.lower()`: this would allow conflicting ES index names for different collections. E.g.: collection `Testdata1` -> index `testdata1`, collection `testdata1` -> index `testdata1`. The same index name for 2 different collections.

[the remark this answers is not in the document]
mugurrus: OK, forced all collection names to contain only lowercase letters.

liquid.py
```diff
@@ -22,9 +26,22 @@
 log.setLevel(LOG_LEVEL)
 
 
+class InvalidCollectionName(RuntimeError):
```
gabriel-v: Python already has an error class for this, ValueError. Please use that, these 10 lines of code are doing nothing special.

liquid.py
```diff
@@ -117,6 +135,10 @@ def get(self, env_key, ini_path, default=None):
 
         return default
 
+    def validate_collection_name(self, name):
```
gabriel-v: This method never uses the `Configuration` object, so it should be a `@classmethod` or a global function.

Also, I think it's simpler to have a `Collection.is_name_valid` classmethod that just checks if the name is correct or not. The caller (in this case, the `initcollection` command) will throw errors and print messages.

liquid.py
```diff
@@ -256,17 +293,49 @@ def alloc(job, group):
...
+    to search.
+    
+    :param name: the collection name
+    """
+    if name not in config.collections:
+        raise RuntimeError('Collection %s does not exist in the liquid.ini file.', name)
+
+    snoop_nomad_service_name = f'snoop-{name}-api'
+
+    try:
```
gabriel-v: Remove this `try/except/catch` because it's only printing the error to `stderr`. The python interpreter will do the same thing.

liquid.py
```diff
@@ -327,16 +396,94 @@ def get_job(hcl_path, substitutions={}):
...
+    collections_dir = Path(config.liquid_volumes) / 'collections'
+    volumes = set([subdir.name for subdir in collections_dir.iterdir() if subdir.is_dir()])
+
+    return volumes - ini_collections
+
+
+def purge_collection(name):
+    """Purge data for the given collection, remove it from hoover search."""
+
+    try:
```
gabriel-v: This passthrough makes no sense here. Please remove the `try`.

liquid.py
```diff
@@ -256,17 +293,49 @@ def alloc(job, group):
...
+    
+    :param name: the collection name
+    """
+    if name not in config.collections:
+        raise RuntimeError('Collection %s does not exist in the liquid.ini file.', name)
+
+    snoop_nomad_service_name = f'snoop-{name}-api'
+
+    try:
+        existing_collections = docker_exec('hoover-search', './manage.py', 'listcollections').split()
```
gabriel-v: Extract this as a global function as you're using it again below.

liquid.py
mugurrus: It makes sense because the exception stacktrace does not contain the output from the command run in the hoover-search service. I just forgot to raise again the exception, which I'll do.

liquid.py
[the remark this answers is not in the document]
mgax: Then please remove the try/except.

liquid.py
mgax: Please don't change the `'1'` here.

liquid.py
```diff
@@ -327,16 +389,94 @@ def get_job(hcl_path, substitutions={}):
...
 
+def gc():
+    """Stop collections jobs that are no longer declared in the ini file."""
+
+    nomad_jobs = nomad.jobs()
+
+    for job in nomad_jobs:
+        if job['ID'].startswith('collection-'):
+            collection_name = job['ID'][len('collection-'):]
+            if collection_name not in config.collections and job['Status'] == 'running':
```
mgax: No need to check for `job['status'] == 'running'`, just stop them all, it won't return an error.

liquid.py
```diff
@@ -285,9 +352,83 @@ def get_job(hcl_path, substitutions={}):
...
+            print(f'Purging collection {coll}...')
+            purge_collection(coll)
+    else:
+        print('No collections will be purged')
+
+
 def deploy():
     """ Run all the jobs in nomad. """
 
+    gc()
```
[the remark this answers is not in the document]
mgax: Sorry to come back to this, but I think `gc` should really be a standalone command. I want us to feel safe running `./liquid deploy` every time.

liquid.py
mugurrus: Yeah, I just don't like seeing stop messages for services that were actually already stopped.

liquid.py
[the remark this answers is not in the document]
mugurrus: Actually I need to process the command output to read the list of collections from the hoover search index. In case something went wrong, I need to print the output of the command in this service. The CalledProcessError exception stacktrace does not contain that data.

liquid.py
mugurrus: The caller is not the `initcollection` command. Also, this method is not supposed to be used outside Configuration, that's why I didn't bother making it a class method.

liquid.py
mugurrus: Nope. The CalledProcessError exception stack trace does not contain the output from the command ran in the hoover search service. In case something went wrong I need to print that output, otherwise I won't know what happened.

liquid.py
```diff
@@ -128,7 +142,8 @@ def run(cmd):
 
 def run_fg(cmd, **kwargs):
     kwargs.setdefault('shell', True)
-    subprocess.check_call(cmd, **kwargs)
+    kwargs.setdefault('stderr', subprocess.STDOUT)
+    return subprocess.check_output(cmd, **kwargs).decode()
```
mgax: Let's specify an encoding. `latin1` is safe for any incoming bytestring and will keep most text as-is.

liquid.py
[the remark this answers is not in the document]
mgax: Then please use another function, perhaps `docker_exec`, and leave `shell` as simple as possible.

liquid.py
```diff
@@ -256,17 +285,50 @@ def alloc(job, group):
...
+
+
+def nomad_address():
+    print(get_nomad_address())
 
 
 def set_collection_defaults(name, settings):
     settings['name'] = name
-    settings.setdefault('workers', '1')
+    settings.setdefault('workers', 1)
```
mgax: And now it's a number instead of a string. Why?

liquid.py
mgax: Fair enough :)

liquid.py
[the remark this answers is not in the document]
mugurrus: But it is `docker_exec`, please take a look at the code, the snippet you see here is obsolete.

liquid.py
mugurrus: My mistake, forgot it has to be a string.

liquid.py
```diff
@@ -256,12 +290,45 @@ def alloc(job, group):
     print(first(running, 'running allocations'))
 
 
-def nomad_address():
+def get_search_collections():
+    try:
+        return run(prepare_command('hoover-search', './manage.py', 'listcollections'), shell=False).split()
```
mgax: Let's extract this bit as a function:
```python
def docker_exec(liquid_task_name, *args):
    return run(prepare_command(liquid_task_name, *args))

docker_exec('hoover-search', './manage.py', 'listcollections')
```

liquid.py
```diff
@@ -327,6 +394,77 @@ def get_job(hcl_path, substitutions={}):
...
+    volumes = set([subdir.name for subdir in collections_dir.iterdir() if subdir.is_dir()])
+
+    return volumes - ini_collections
+
+
+def purge_collection(name):
+    """Purge data for the given collection, remove it from hoover search."""
+
+    if name in get_search_collections():
+        shell('hoover-search', './manage.py', 'removecollection', name)
```
mgax: Again, let's use `search` only as a command.

liquid.py
```diff
@@ -256,12 +290,45 @@ def alloc(job, group):
...
+    :param name: the collection name
+    """
+    if name not in config.collections:
+        raise RuntimeError('Collection %s does not exist in the liquid.ini file.', name)
+
+    if name in get_search_collections():
+        log.info(f'Collection "{name}" was already initialized.')
+        return
+
+    shell(f'snoop-{name}-api', './manage.py', 'initcollection')
```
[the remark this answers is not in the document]
mugurrus: `shell` prints the command output, including errors, which is what we need here. Otherwise I would need to write again that try-except sequence.

liquid.py
mugurrus: At least for now, it's only used once and for combining 2 function calls in one line I think this is excessive. Regarding `shell`, we need it because it needs to be registered as a command.

liquid.py
mugurrus: Same as above

liquid.py
```diff
@@ -233,14 +252,29 @@ def first(items, name_plural='items'):
...
+
+    :param name: the value of the liquid_task tag
+    :param tty: if true, instruct docker to allocate a pseudo-TTY and keep stdin open
     """
     containers = docker.containers([('liquid_task', name)])
     container_id = first(containers, 'containers')
-    docker_exec_cmd = ['docker', 'exec', '-it', container_id] + list(args or ['bash'])
-    run_fg(docker_exec_cmd, shell=False)
+
+    args = [arg.replace('__nomad_address__', get_nomad_address()) for arg in args]
```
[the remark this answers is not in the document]
mugurrus: It was being used in the shell command to add the collection to hoover search. I think it could still be useful in the future.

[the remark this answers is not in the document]
mgax: Then we can add it when we need it - [YAGNI](https://en.wikipedia.org/wiki/You_aren't_gonna_need_it).

liquid.py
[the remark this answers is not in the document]
mgax: Then move the try/except in `docker_exec`.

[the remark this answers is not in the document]
mgax: Ok fine let's just call `shell` here and let the user see the output as-is.

PR #33, opened 2019-04-25
split liquid.py into separate modules

mugurrus:


hack.md
```diff
@@ -0,0 +1,90 @@
+# CLI commands test procedure
```
gabriel-v: Rename this as `docs/testing.md`

PR #34, opened 2019-04-25
Vagrant update

alexneamtu:
- Updated vagrant file to follow the instructions in https://hackmd.io/MeBET7lQSd2mBOv7vlhwWw


Vagrantfile
```diff
@@ -15,6 +15,4 @@ Vagrant.configure("2") do |config|
 
   config.vm.provision :shell, path: "vagrant-provision.sh"
 
-  config.vm.synced_folder ".", "/vagrant"
```
mgax: We still need some kind of synched folder, and on the `libvirt` backend, this was the only way. If it's only useful for `libvirt` then we can hide it behind an `if`.

mgax: Eh never mind, @gabriel-v says it's usually not reliable.

vagrant/provision-cluster.sh
```diff
@@ -0,0 +1,42 @@
...
+cd /opt/cluster
+
+cat > cluster.ini <<EOF
+[nomad]
+address = 0.0.0.0
+[vault]
+disable_mlock = true
+EOF
+
+./cluster.py install
```
gabriel-v: `sudo setcap cap_ipc_lock=+ep bin/vault`

vagrant/provision-cluster.sh
```diff
@@ -0,0 +1,42 @@
...
+echo "Installing the nomad cluster"
+
+git clone https://github.com/liquidinvestigations/cluster /opt/cluster
+cd /opt/cluster
+
+cat > cluster.ini <<EOF
+[nomad]
+address = 0.0.0.0
+[vault]
+disable_mlock = true
```
gabriel-v: `false`

vagrant/provision-cluster.sh
```diff
@@ -0,0 +1,42 @@
...
+EOF
```
gabriel-v:
```
[supervisor]
autostart = true
```

So that following `vagrant up` calls will have everything running.

vagrant/provision-cluster.sh
```diff
@@ -0,0 +1,42 @@
...
+
+./cluster.py install
+./cluster.py configure
+sudo ln -s `pwd`/etc/supervisor-cluster.conf /etc/supervisor/conf.d/cluster.conf
+sudo supervisorctl update
+
+sudo supervisorctl start cluster:consul cluster:vault
+./cluster.py autovault
+sudo supervisorctl start cluster:nomad
```
gabriel-v: `restart` because you'll have `autostart = true` and it will die.

vagrant/provision-cluster.sh
```diff
@@ -0,0 +1,42 @@
...
+[vault]
+disable_mlock = true
+EOF
+
+./cluster.py install
+./cluster.py configure
+sudo ln -s `pwd`/etc/supervisor-cluster.conf /etc/supervisor/conf.d/cluster.conf
+sudo supervisorctl update
+
+sudo supervisorctl start cluster:consul cluster:vault
```
gabriel-v: Unneeded because of `autostart = true`.

PR #48, opened 2019-04-27
Manage the venv

mgax:
Auto-recreate the venv if it's out of date, and provide a sane path to add new dependencies and update their versions.

### DIFF
--- a/liquid
+++ b/liquid
@@ -4,10 +4,16 @@ set -e
 # cd to file
 cd "$(dirname ${BASH_SOURCE[0]})"
 
+# delete venv if outdated
+if ! shasum -c venv/requirements.shasum; then
+  rm -rf venv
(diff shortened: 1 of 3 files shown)


liquid
```diff
@@ -4,10 +4,16 @@ set -e
 # cd to file
 cd "$(dirname ${BASH_SOURCE[0]})"
 
+# delete venv if outdated
+if ! shasum -c venv/requirements.shasum; then
```
gabriel-v: `if ! shasum -c venv/requirements.shasum requirements.txt; then` right?

mgax: No, `requirements.shasum` contains checksums and filenames.  Well, a single checksum, and a single filename. But that filename is exactly `requirements.txt`. :)

PR #55, opened 2019-04-30
Jinja2

gabriel-v:


liquid_node/jobs.py
```diff
@@ -79,6 +80,17 @@ def get_collection_job(name, settings, template='collection.nomad'):
     return get_job(config.templates / template, substitutions)
 
 
+
+
+def render(template, subs):
+    print(subs)
```
mgax: I'm ok with two blank lines above but pls no print.

PR #60, opened 2019-05-02
DokuWiki integration

mgax:
Integrate DokuWiki into the cluster. Authentication is handled by the trusty old authproxy and a 30-line dokuwiki plugin, mainly because I want to make sure no unauthenticated user can possibly see any of the wiki.

This PR also extracts authproxy as a reusable jinja macro. Woo templating!


liquid_node/commands.py
```diff
@@ -108,13 +113,15 @@ def deploy():
 
     consul.set_kv('liquid_domain', config.liquid_domain)
     consul.set_kv('liquid_debug', 'true' if config.liquid_debug else 'false')
+    consul.set_kv('liquid_http', config.liquid_http)
```
gabriel-v: Let's name these `liquid_http_protocol`or something similarly specific. Maybe we could use a `liquid_public_url` entry in the consul kv instead of that `${liquid_http}://{{key "liquid_domain"}}` that keeps appearing in this patch.

templates/_lib.hcl
```diff
@@ -1,9 +1,69 @@
...
     attempts = 10
     interval = "5m"
     delay = "5s"
     delay_function = "exponential"
     max_delay = "30s"
   }
-{% endmacro %}
+{%- endmacro %}
+
+{%- macro authproxy(name, host, upstream) %}
```
gabriel-v: Maybe rename this to `authproxy_group(..)` to keep the naming scheme?

mgax: Yup great idea thanks :+1:

PR #67, opened 2019-05-13
Rocket.Chat

mgax:
* [x] Document the manual step (described below)
* [ ] ~Make sure the OAuth2 integration survives a realloc of the `liquid-core` task on a different port~ we'll fix it when it happens
* [x] Set a secret admin password
* [x] Deploy on liquiddemo.org
* [ ] ~On first login, users can pick a username, any username. We need to pin the username to the liquid-core username.~ Or maybe we don't need this.

Deploy _Rocket.Chat_ on the cluster and configure a custom OAuth2 provider. There is a single manual step: this Meteor method must be called somehow, and the easiest method is through the UI.

https://github.com/RocketChat/Rocket.Chat/blob/d53e264b83ccc971236123853e4b68c9eb6beeb4/app/lib/server/methods/addOAuthService.js#L11

Screenshot from 2019-05-14 01-41-22

Screenshot from 2019-05-14 01-41-53

The OAuth config values are already present in the task's environment and they will get picked up automatically.

Screenshot from 2019-05-14 01-45-31


mgax:
> Please add the instructions and merge this as it is.

Here's some docs: https://github.com/liquidinvestigations/node/blob/893fa132a1576fe16b7636ba119b3af7050f2fe6/docs/RocketChatAuthSetup.md

PR #69, opened 2019-05-16
Nextcloud: 1st steps integrating auth provider & TLS support added

salevajo:


mgax:
@salevajo thanks for this! I expect that's all we need to do. Does it work? Do you want to make more changes or can we just merge it? If it's ready, mark it as "ready for review", and I'll pick it up from there.

salevajo:
I don't think it's ready, I'll assign @MattesR. I talked about that with him briefly on Monday.

MattesR:
From looking at it, it is exactly what I would've done myself, I can test it (probably tomorrow)
@mgax how can I best test whether the auth proxy does the right thing? Not being logged in on liquid and trying to access nextcloud?

mgax:
> how can I best test whether the auth proxy does the right thing? Not being logged in on liquid and trying to access nextcloud?

Yes. You can also check the `/__auth/logout` view, it should always say "you have been logged out", served directly by authproxy.

MattesR:
I've checked out the PR and there are still errors which I was unable to fix so far. I don't see any difference between the implementation of the authproxy for nextcloud vs the implementation for dokuwiki, however it is not working for nextcloud. The authproxy restarts over and over again.
After restarting due to an alleged change in the template which I couldn't confirm (I've checked out `/local/docker.env` before and after rerendering, nothing changed) the container exits with error code 137 which indicates a OOM sigkill. I've upped the allocated memory, which was never depleted in the first place but the container still terminates with the same error code.

MattesR:
I've checked out the master branch and apllied the changes of this PR again and I still get the same errors. Nextcloud is (with or without authproxy) only available on the port that nomad assigns to it so `nextcloud.<liquid_domain>` will result in a 404 message (plain `404 page not found`) but `nextcloud.<liquid_domain>:PORT_FROM_NOMAD` will get you to nextcloud. 
I'm not sure how to fix this yet. It seems to be a networking issue? I'm also still confused by the fact that the authproxy is always restarting.

mgax:
> I'm also still confused by the fact that the authproxy is always restarting.

That must be the problem. When it will start and stay up, it will be marked as "healthy" in consul, so traefik will begin to route traffic to it.

PR #71, opened 2019-05-17
import collections from docker-setup

mugurrus:


liquid.py
```diff
@@ -42,6 +42,7 @@ def main():
         commands.initcollection,
         commands.purge,
         commands.getsecret,
+        commands.importfromdocker
```
mgax: Final comma pls.

liquid_node/import_from_docker.py
```diff
@@ -0,0 +1,107 @@
...
+        src.replace(dst)
+    elif method == 'link':
+        dst.symlink_to(src)
+
+
+def import_index(docker_setup, method):
+    log.info('Importing hoover search datadabase')
+    pg_src = docker_setup / 'volumes' / 'search-pg'
+    pg_dst = Path(config.liquid_volumes) / 'hoover' / 'pg'
+    if pg_dst.is_symlink():
```
mgax: This feels a bit drastic. I think we should bail if the target folder exists.

liquid_node/import_from_docker.py
```diff
@@ -0,0 +1,107 @@
...
+    if method  not in import_methods:
+        raise RuntimeError(f'Invalid import method {method}')
+
+    if dst.is_dir():
+        shutil.rmtree(str(dst))
+    elif dst.is_symlink():
+        dst.unlink()
+
+    if method == 'copy':
+        shutil.copytree(str(src), str(dst))
```
mgax: We should `log.info` these steps.

templates/hoover-migrate.nomad
```diff
@@ -29,7 +29,7 @@ job "hoover-migrate" {
             SECRET_KEY = {{.Data.secret_key}}
           {{- end }}
           {{- range service "hoover-pg" }}
-            HOOVER_DB = postgresql://hoover:hoover@{{.Address}}:{{.Port}}/hoover
+            HOOVER_DB = postgresql://search:search@{{.Address}}:{{.Port}}/search
```
mgax: Yep. We agreed to rename the pg user and database to match the ones from docker-setup. Because it makes the migration easier and why not.

liquid.py
```diff
@@ -42,6 +42,7 @@ def main():
         commands.initcollection,
         commands.purge,
         commands.getsecret,
+        commands.importfromdocker,
```
mgax: Let's call it `importfromdockersetup` because technically we're still running docker.

liquid_node/commands.py
```diff
@@ -294,6 +299,44 @@ def purge(force=False):
         print('No collections will be purged')
 
 
+def importfromdocker(path, method='link'):
```
mgax: Let's move all of this logic to `import_from_docker.py`, it's conceptually isolated.

liquid_node/import_from_docker.py
```diff
@@ -0,0 +1,108 @@
...
+            names[name.lower()].append(name)
+        else:
+            names[name] = [name]
+    if conflicting:
+        groups = '; '.join([', '.join(names[conflict]) for conflict in conflicting])
+        raise RuntimeError(f'These collections have conflicting names: {groups}. '
+                           'Rename the conflicting collections in docker-setup.')
+
+
+import_methods = ['copy', 'move', 'link']
```
mgax: Nice :)

liquid_node/commands.py
mugurrus: It's a command; there are different commands for different functionality already in `commands.py`.  In my opinion it should stay here.

mgax: Then move most of the implementation in the other module, there's no reason to import 5 functions instead of just one.

mugurrus: The number of imported functions is irrelevant. `deploy` also uses a many functions, and some of them are in `commands.py`, although they should not be there.

mgax: Yep, deploy should probably be its own module. So let's not clutter `commands.py` any more than we have to.

mugurrus: My opinion is that `importfromdockersetup`, as a command, is in the right module. There are other functions that indeed should not be there, but it's not this function that clutters the space.

mgax: Then keep the function and move all the logic to a new function in `import_from_docker`?

PR #75, opened 2019-05-27
add LIQUID_DOMAIN env for liquid

MattesR:


MattesR:
I think the `SERVICE_ADDRESS` also has to be set, right?!
I've set it on my machine by looking up the address first but how can we make it in general?

mgax:
This is covered in https://github.com/liquidinvestigations/node/pull/74 actually.

MattesR:
I'll close it then

PR #95, opened 2019-06-09
Nextcloud healthcheck

mgax:
Use `/status.php` for the healthcheck. Not sure if it actually works tho.


gabriel-v:
stole your code a while ago

PR #96, opened 2019-06-12
development repositories mount fixes

mugurrus:
- allow relative paths (previously the mount paths had to be absolute)
- log warning when a repository path did not exist and do not mount it

### DIFF
--- a/liquid_node/jobs.py
+++ b/liquid_node/jobs.py
@@ -1,5 +1,6 @@
 import logging
 import os
+from pathlib import Path
 
 import jinja2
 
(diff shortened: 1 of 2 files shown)


liquid_node/jobs.py
```diff
@@ -69,13 +70,14 @@ def set_volumes_paths(substitutions={}):
...
 
+        substitutions[key] = ''
         if config.mount_local_repos:
-            substitutions[key] = repo_volume_line
-        else:
-            substitutions[key] = ''
+            if Path(repo_config['local']).is_dir():
+                substitutions[key] = f"\"{repo_config['local']}:{repo_config['target']}\",\n"
+            else:
+                log.warn(f'Invalid repo path "{repo_config["local"]}"')
```
gabriel-v: Crash here with a `RuntimeError` or `ValueError`. This is not a warning, this is the end of the universe.

mugurrus: The warning is there on purpose, to allow only part of the repositories to be mounted.

PR #98, opened 2019-06-13
WIP: Change the initial admin username for nextcloud

gabriel-v:
Check this patch manually before merging! Right now it lands me into a redirect loop.

After merging this we're going to have to instruct everybody to `./liquid halt`, `sudo rm -rf /opt/volumes/nextcloud`, `./liquid deploy`


gabriel-v:
> Check this patch manually before merging! Right now it lands me into a redirect loop.
False alarm, this should just work

gabriel-v:
Will also need to always the admin user password when the job starts, because before this patch it used to be `admin`.

PR #101, opened 2019-06-16
User-set titles for liquid-core, rocketchat, nextcloud and dokuwiki

gabriel-v:
rocket
doku

Nothing for Hoover at the moment...
hoover

see https://github.com/liquidinvestigations/liquid-nextcloud/pull/2
and https://github.com/liquidinvestigations/core/pull/49


examples/liquid.ini
```diff
@@ -11,6 +11,11 @@ consul_socket = ../cluster/var/consul.socket
 # Domain for the liquid bundle.
 domain = liquid.example.org
 
+# User-friendly title.
+# Defaults to the domain with no spaces, capitalized.
```
mgax: How about we add more magic and remove the last bit of the domain name? That would actually yield more relevant vaues.

templates/nextcloud-migrate.nomad
```diff
@@ -15,7 +15,7 @@ job "nextcloud-migrate" {
 
       driver = "docker"
       config = {
-        image = "liquidinvestigations/liquid-nextcloud"
+        image = "liquidinvestigations/liquid-nextcloud:title"
```
mgax: Don't forget to remove these.

templates/rocketchat.nomad
```diff
@@ -69,16 +69,16 @@ job "rocketchat" {
             ADMIN_PASS={{.Data.pass}}
           {{- end }}
           ADMIN_EMAIL=admin@example.com
-          Organization_Name=Liquid
-          Site_Name=Liquid
+          Organization_Name=${config.liquid_title} Liquid
```
mgax: Why keep a trailing "Liquid" here?

gabriel-v: removed

templates/nextcloud-migrate.nomad
gabriel-v: done

examples/liquid.ini
mgax: overkill

PR #107, opened 2019-06-17
Deployment stats

gabriel-v:
Print memory, disk IO and ram required to deploy.

### DIFF
--- a/liquid_node/commands.py
+++ b/liquid_node/commands.py
@@ -1,4 +1,5 @@
 from pathlib import Path
+from collections import defaultdict
 from time import time, sleep
 import logging
 import os
@@ -132,6 +133,26 @@ def wait_for_service_health_checks(health_checks):
     raise RuntimeError(msg)
 
 
+def resources():
+    """Get memory and CPU usage for the deployment"""
+
+    def get_all_res():
+        jobs = [nomad.parse(get_job(job.template)) for job in config.jobs]
+        for spec in jobs:
+            yield from nomad.get_resources(spec)
+
+    total = defaultdict(int)
+    for name, _type, res in get_all_res():
+        for key in ['MemoryMB', 'CPU']:
+            if res[key] is None:
+                raise RuntimeError("Please update Nomad to 0.9.3+")
+            total[f'{_type} {key}'] += res[key]
+
+    print('Resource requirement totals: ')
+    for key, value in sorted(total.items()):
+        print(f'  {key}: {value}')
+
+
 def deploy():
     """Run all the jobs in nomad."""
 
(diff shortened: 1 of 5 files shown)


liquid_node/commands.py
```diff
@@ -1,4 +1,5 @@
 from pathlib import Path
+import collections
```
mgax: Please write this as `from collections import defaultdict`, we already have the concept of collections, so this would avoid confusion.

liquid_node/commands.py
```diff
@@ -137,6 +138,25 @@ def get_failed_checks():
...
+
+    def get_all_res():
+        jobs = (nomad.parse(get_job(path)) for _, path in config.jobs)
+        for spec in jobs:
+            yield from nomad.get_resources(spec)
+
+    total = collections.defaultdict(int)
+    for _, _type, res in get_all_res():
+        for key in ['MemoryMB', 'CPU', 'DiskMB']:
+            totalkey = _type + " " + key
```
mgax: how 'bout `f'{_type} {key}'` and merge with the next line? It's even more readable that way.

liquid_node/nomad.py
```diff
@@ -21,12 +23,34 @@ def name(check):
...
+                res = {}
+            for key in DEFAULTS:
+                if not res.get(key):
+                    res[key] = DEFAULTS[key]
+            return res
+
+        for group in spec['TaskGroups'] or []:
+            for task in group['Tasks'] or []:
+                name = f'{spec["Name"]}-{group["Name"]}-{task["Name"]}'
+                yield name, spec['Type'], set_defaults(task['Resources'])
```
mgax: `set_defaults(task['Resources'])` is like `resources = dict(DEFAULTS); resources.update(task['Resources'] or {})`

liquid_node/nomad.py
```diff
@@ -21,12 +23,34 @@ def name(check):
...
         for group in spec['TaskGroups'] or []:
             for task in group['Tasks'] or []:
                 for service in task['Services'] or []:
                     yield service['Name'], [name(check) for check in service['Checks'] or []]
 
+    def get_resources(self, spec):
+        """Generates (task, type, resources) tuples with resource stranzas for
+        the supplied job, with default values set for memory, disk and cpu."""
+
+        DEFAULTS = {
```
mgax: Can we get this from nomad? They could differ depending on how it's built/configured.

mgax: Even better: https://www.nomadproject.io/api/jobs.html#canonicalize and forget about the next comment

vagrant/provision-liquid.sh
```diff
@@ -31,6 +31,7 @@ fi
 
 cp examples/liquid.ini .
 
+./liquid deploystats
```
mgax: So this prints the total resource needs of the deployment, not any real stats. Let's call it `resources` or something.

PR #108, opened 2019-06-17
Use a newer vmck image and configure resources

mgax:


templates/drone.nomad
```diff
@@ -111,6 +111,8 @@ job "drone" {
         {{- range service "vmck-imghost" }}
           QEMU_IMAGE_URL = "http://{{.Address}}:{{.Port}}/cluster-master.qcow2.tar.gz"
         {{- end }}
+        QEMU_MEMORY = 12000
+        QEMU_CPU = 2000
```
gabriel-v: This is too slow and things might timeout with high load; let's set it at 7000.

templates/drone.nomad
```diff
@@ -71,7 +71,7 @@ job "drone" {
 
       driver = "docker"
       config {
-        image = "vmck/vmck:0.0.1-permit-archived-artifact"
+        image = "vmck/vmck:configure-resources"
```
gabriel-v: you can use the latest tag now.

templates/drone.nomad
mgax: Well, this is the minimum guaranteed CPU, not a cap on CPU usage. The system will let qemu max out the physical CPU. I'd rather keep the limit low, so we don't get surprise allocation errors, when there is plenty of memory but not so much CPU.

PR #110, opened 2019-06-18
Define jobs like models

mgax:
The jobs code is getting out of hand. I think it's time to refactor.
* [x] Create a class for each of our fixed jobs, subclassed from `Job`, like a model
* [x] Move vault secret management to the Job class
* ~~Convert collections to job models~~ (later)
* ~~Move hcl rendering to a `Job.hcl()` method~~ (later)
* ~~Move volume path arithmetic to the Job class~~ (later)

This PR works as-is, so maybe we merge it first, and do the next steps later. Thoughts?


examples/litterbox.py
```diff
@@ -0,0 +1,7 @@
+from pathlib import Path
+
+
+class Cat
```
gabriel-v: syntax error

examples/liquid.ini
```diff
@@ -72,6 +72,10 @@ wait_green_count = 10
 [job:zipkin]
 template = templates/zipkin.nomad
 
+# Run a custom local job, see litterbox.py
```
gabriel-v: We need to test this config on the CI.

We can either put it in `vagrant/provision-liquid.sh`, but that will also taint the virtualbox VMs with proverbial cats and litterboxes.

If you don't want to do that now, please open an issue for it.

liquid_node/commands.py
```diff
@@ -147,18 +142,23 @@ def deploy():
...
+        'liquid/authdemo/auth.django',
+        'liquid/nextcloud/nextcloud.admin',
+        'liquid/nextcloud/nextcloud.maria',
+        'liquid/dokuwiki/auth.django',
+        'liquid/nextcloud/auth.django',
+        'liquid/rocketchat/auth.django',
+        'liquid/ci/vmck.django',
+        'liquid/ci/drone.secret',
     ]
+    core_auth_apps = list(CORE_AUTH_APPS)
```
mugurrus: This member should be named `auth_apps`, since the list will potentially contain apps that are not core.

liquid_node/jobs/__init__.py
```diff
@@ -120,3 +125,8 @@ def get_job(hcl_path, substitutions={}):
     output = render(template, set_volumes_paths(substitutions))
     log.debug(f'{hcl_path}\n{output}')
     return output
+
+
+class Job:
+    vault_secret_keys = ()
+    core_auth_apps = ()
```
mugurrus: Name should be `auth_apps`

examples/liquid.ini
mugurrus: For non-devs it would be easier if the job was defined in ini. E.g.:
```[job:Cat]
template = ...
```

gabriel-v: that still works https://github.com/liquidinvestigations/node/pull/110/files#diff-bacec7d196e27ee42066f951fd810473R130

mgax: Ok, I've made an issue. https://github.com/liquidinvestigations/node/issues/118

liquid_node/commands.py
mgax: Well, the list contains *only* apps that are not core, because they authenticate against core. Hence the variable name: "apps that do auth with core".

Issue #109, opened 2019-06-17
Command to delete a collection


mgax:
Maybe I have too many collections, or I want to delete one and re-create it from scratch. Here are the manual steps:

* `sudo rm -rf volumes/collections/testdata`
* `./liquid shell hoover-es curl -XDELETE localhost:9200/testdata`
* Delete the collection from hoover-search admin UI

PR #114, opened 2019-06-20
`deletecollection` command

mgax:
Fixes https://github.com/liquidinvestigations/node/issues/109

### DIFF
--- a/liquid_node/commands.py
+++ b/liquid_node/commands.py
@@ -337,6 +337,12 @@ def purge(force=False):
         print('No collections will be purged')
 
 
+def deletecollection(name):
+    """Delete a collection by name"""
+    nomad.stop(f'collection-{name}')
+    purge_collection(name)
+
+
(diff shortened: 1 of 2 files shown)


liquid_node/commands.py
```diff
@@ -337,6 +337,12 @@ def purge(force=False):
         print('No collections will be purged')
 
 
+def deletecollection(name):
+    """Delete a collection by name"""
+    nomad.stop(f'collection-{name}')
+    purge_collection(name)
```
mugurrus: This will not remove it from the ini file

mgax: And it should not. I see two scenarios:
1. Really remove a collection - run `deletecollection` and remove it from the ini file, in any order.
2. Reprocess a collection from scratch, useful in development and testing - run `deletecollection` and then `deploy`.

I'm mostly interested in the 2nd scenario, and with this command, iterating is pretty smooth.

PR #116, opened 2019-06-20
fir gnupg error on snoop indexing

mugurrus:
for task https://github.com/CRJI/EIC/issues/230

### DIFF
--- a/templates/collection.nomad
+++ b/templates/collection.nomad
@@ -139,6 +139,7 @@ job "collection-${name}" {
         SNOOP_COLLECTION_ROOT = "collection"
         SNOOP_TASK_PREFIX = "${name}"
         SNOOP_ES_INDEX = "${name}"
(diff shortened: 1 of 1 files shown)


mgax:
This makes the error go away, but it doesn't actually process any PGP emails, because nobody is setting up keys, right?

I think the way to fix the bug is to change the contract of snoop: it should have `gpghome` in the collection folder (next to `data` and `ocr` we define another folder) already created and populated with keys. If there is no `gpghome` then no attempt is made to process PGP email parts.

gabriel-v:
Closing this patch as it's not fixing the problem. We will use a per-collection `gpghome` directory as laid out in https://github.com/hoover/snoop2/pull/167. Thanks for pointing out the duplication that is happening in https://github.com/liquidinvestigations/node/blob/908e0d9cbb10fcef6f3b850932ba04b6d189fb8e/templates/collection.nomad#L201.

PR #123, opened 2019-06-27
importfromdockersetup: Don't overwrite ini file; link collections too

gabriel-v:
Simplify migrating:

- don't overwrite liquid.ini, just ask the user to edit it
- link/copy/move the collection source data too


gabriel-v:
It was your good idea.

PR #124, opened 2019-06-27
wait for env vars initialisation in the collection services

mugurrus:
The snoop api and workers services keep being restarted until the environment var are set (after the corresponding services started). This script waits for the env vars to be set without exiting, reduces the processor usage.

### DIFF
--- a/templates/_lib.hcl
+++ b/templates/_lib.hcl
@@ -91,3 +91,24 @@ ephemeral_disk {
     }
   }
 {%- endmacro %}
+
+
+{%- macro wait_env_vars_script(vars, command, attempts=10, delay=10) %}
+  data = <<-EOF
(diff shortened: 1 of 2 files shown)


mugurrus:
1, 2 answered in the description.
3: This PR was not meant for "change mid-allocation"

mgax:
> 3: This PR was not meant for "change mid-allocation"

The loop will continually check for the env variables in the same shell process. Those will never change while the process is running. And, in any case, Nomad will restart the container if any env var changes. So you could simplify the script to just sleep if any var is missing.

Now, it's happened that Nomad fails to restart a container when a template should be re-rendered, maybe because it's still v0.x. I think that, for now, it's best to just let the containers restart until they work.

PR #125, opened 2019-06-27
allow only hoover jobs to be started

mugurrus:
For dev purposes I need to be able to start LI faster, without using so many resources. When working on search and snoop it's enough to start only liquid, hoover and collections jobs.

### DIFF
--- a/liquid_node/commands.py
+++ b/liquid_node/commands.py
@@ -24,7 +24,7 @@ from .import_from_docker import import_collection
 
 log = logging.getLogger(__name__)
 
-CORE_AUTH_APPS = [
+HOOVER_AUTH_APPS = [
     {
         'name': 'authdemo',
         'vault_path': 'liquid/authdemo/auth.oauth2',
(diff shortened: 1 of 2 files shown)


liquid_node/commands.py
```diff
@@ -167,15 +170,18 @@ def deploy():
...
-        'liquid/nextcloud/nextcloud.admin',
-        'liquid/nextcloud/nextcloud.maria',
-        'liquid/dokuwiki/auth.django',
-        'liquid/nextcloud/auth.django',
-        'liquid/rocketchat/auth.django',
-        'liquid/ci/vmck.django',
-        'liquid/ci/drone.secret',
     ]
-    core_auth_apps = list(CORE_AUTH_APPS)
+    if scope != 'hoover':
```
mgax: I don't think it's wise to special-case hoover, it should be just another app that we deploy.

liquid_node/commands.py
```diff
@@ -208,7 +214,8 @@ def start(job, hcl):
             job_checks[service] = checks
         return job_checks
 
-    jobs = [(job.name, get_job(job.template)) for job in config.jobs]
+    config_jobs = config.jobs if scope != 'hoover' else config.hoover_jobs
```
mgax: This is exactly the kind of `if` that will make the code unmaintainable in no time. If you only want to deploy hoover, then @gabriel-v 's idea in https://github.com/liquidinvestigations/node/pull/125#pullrequestreview-255718556 makes more sense.

PR #126, opened 2019-06-27
fix crash on empty response

mugurrus:
https://github.com/CRJI/EIC/issues/266

### DIFF
--- a/liquid_node/jsonapi.py
+++ b/liquid_node/jsonapi.py
@@ -40,8 +40,10 @@ class JsonApi:
 
         with urlopen(req) as res:
             if res.status == 200:
-                res_body = json.load(res)
-                return res_body
+                res_body = res.read()
+                if len(res_body) == 0:
+                    return None
+                return json.loads(res_body)
 
     def get(self, url):
         return self.request('GET', url)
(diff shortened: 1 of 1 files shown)


liquid_node/jsonapi.py
```diff
@@ -40,8 +40,10 @@ def request(self, method, url, data=None, headers=None):
 
         with urlopen(req) as res:
             if res.status == 200:
-                res_body = json.load(res)
-                return res_body
+                res_body = res.read()
+                if len(res_body) == 0:
```
mgax: Maybe instead check the Content-Type header? If it's `application/json` then parse it as json.

mugurrus: The Content-Type header does not exist in the response from the nomad URL that's returning empty content. On the other hand, if the response was not JSON, the JSON parser will raise an exception, so this case is treated already.
It's an inconsistency on their part (nomad) that in a REST API on one URL the response is empty content with no Content-Type. But I need to treat that as well, otherwise it crashes in liquid.

mgax: That's ok, if there is no Content-Type, then it's clearly not JSON. So you can do `res.headers.get('Content-Type') == 'application/json'` and it will handle all cases.

gabriel-v:
Now it seems you're breaking everything else: 

    drone-397: 2019-06-28 12:43:56 ERROR HTTP Error <HTTPError 204: 'No Content'>: b''

mgax:
:ship: cool!

PR #127, opened 2019-06-27
cleanup dead jobs from nomad

mugurrus:
Remove the dead jobs from the jobs list in nomad
Requires https://github.com/liquidinvestigations/node/pull/126


liquid_node/commands.py
```diff
@@ -246,14 +246,26 @@ def halt():
 def gc():
     """Stop collections jobs that are no longer declared in the ini file."""
 
-    nomad_jobs = nomad.jobs()
-
-    for job in nomad_jobs:
+    killed_jobs = []
```
gabriel-v: `jobs_to_stop`?

liquid_node/commands.py
```diff
@@ -246,14 +246,26 @@ def halt():
...
+    while killed_jobs:
+        sleep(config.wait_interval)
+
+        nomad_jobs = {job['ID']: job for job in nomad.jobs() if job['ID'] in killed_jobs}
+        for job_name in killed_jobs:
+            if job_name not in nomad_jobs or nomad_jobs[job_name]['Status'] == 'dead':
+                killed_jobs.remove(job_name)
+                log.info(f'Job {job_name} is dead')
+
+    nomad.gc()
```
gabriel-v: Is this really useful in our case? This removes the logs and past failures and they might be useful when debugging something and adding/removing collectitons.

liquid_node/commands.py
mugurrus: The list will contain jobs that were killed in the loop, not jobs that are to be stopped after the loop.

liquid_node/commands.py
```diff
@@ -298,17 +298,33 @@ def halt():
...
+    log.info(f'Waiting for jobs to die...')
+    timeout = time() + config.wait_max
+    while killed_jobs and time() < timeout:
+        sleep(config.wait_interval)
+
+        nomad_jobs = {job['ID']: job for job in nomad.jobs() if job['ID'] in killed_jobs}
+        for job_name in killed_jobs:
+            if job_name not in nomad_jobs or nomad_jobs[job_name]['Status'] == 'dead':
+                killed_jobs.remove(job_name)
+                log.info(f'Job {job_name} is dead')
```
gabriel-v: Please thow an error here if jobs are still alive:
```python3
while ...:
    ...
    if not killed_jobs:
        break
else:
    raise RuntimeError('jobs are still up: ' + str(killed_jobs))

File: liquid_node/commands.py (hunk 1 above)
Comment (gabriel-v) [reply liquid_node/commands.py]: The nomad command is called `stop`, not `kill`. There is no killing or force stopping happening here.

Also, when you declare the list, nothing has been stopped yet. Having a variable called `*_killed` or `*_stopped` implies those things have already been `killed`/`stopped`, this is not the case here.

A name like `jobs_to_die` would make it clear that the jobs are still alive at line 304.

File: liquid_node/commands.py [hunk 4]
```diff
@@ -298,17 +298,33 @@ def halt():
...
+
+
+def cleandeadjobs():
```
gabriel-v: A clearer name for this might be `nomad_gc`, since it's just calling that method.

liquid_node/commands.py
mugurrus: Again, the list is meant to store jobs that were stopped. 304 is the initialisation point, but the meaning of a variable is not in general given by the initialisation point. I'll change `killed` to `stopped`.

liquid_node/commands.py
mugurrus: There are situations when the batch jobs ran OK, and I don't care about them, I just need to see the jobs that I'm working on without any clutter.

liquid.py
```diff
@@ -37,7 +37,8 @@ def main():
         commands.nomad_address,
         commands.deploy,
         commands.resources,
-        commands.gc,
+        commands.cleancollectionjobs,
```
gabriel-v: I think we should keep the `gc` command named as it was, and have it also remove extra collection jobs. The `nomadgc` addition is really useful here through.

mugurrus: This may confusion, as gc, which is generic, doesn't garbage collect all dead jobs, only the collections ones. Maybe `collectionsgc`?

Readme.md
```diff
@@ -247,3 +247,13 @@ template = local/foo.nomad
 ```
 
 Afterwards, run `./liquid deploy`, which will send your job `foo` to nomad.
+
+#### Removing collections
+In order to remove a collection, take the following steps:
+1. Remove the corresponding collection section from the `liquid.ini` file.
+2. Run `./liquid cleancollectionjobs`
```
gabriel-v: this has been renamed

PR #128, opened 2019-06-28
Db in vault

MattesR:


MattesR:
I didn't put the databases into separate job, as you suggested @gabriel-v, since we can check for db related health checks without it.
Since there was more to do than just run a psql command, I put everything into a script and place it inside the container.

templates/collection.nomad
```diff
@@ -92,9 +92,39 @@ job "collection-${name}" {
...
       }
-      env {
-        POSTGRES_USER = "snoop"
-        POSTGRES_DATABASE = "snoop"
+      template {
+        data = <<EOF
+          POSTGRES_USER = "snoop"
+          POSTGRES_DATABASE = "snoop"
+          {{- with secret "liquid/collections/${name}/snoop.postgres" }}
+            POSTGRES_PASSWORD = {{.Data.secret_key}}
```
gabriel-v: Shouldn't this be in quotes like the ones above it? I guess the `env = true` files get parsed by some golang library.

If we have been using this with and without them around the code base everywhere, we probably really need to fix that. Could you put quotes around everything in your patch?

Thanks!

templates/hoover.nomad
```diff
@@ -63,9 +63,39 @@ job "hoover" {
...
+        then
+          psql -U search -c "ALTER USER search password '$POSTGRES_PASSWORD'"
+          sed -i '$d' $PGDATA/pg_hba.conf
+          sed -i '$d' $PGDATA/pg_hba.conf
+          {
+            echo
+            echo "host all all all md5"
+            echo
+          } >> "$PGDATA/pg_hba.conf"
+          echo changed password of database
```
gabriel-v: `database password changed`?

liquid_node/commands.py
```diff
@@ -217,6 +219,13 @@ def start(job, hcl):
         job_checks = start(job, hcl)
         health_checks.update(job_checks)
 
+    # Wait for database health checks of all collections and hoover:
+    pg_checks = {k: v for k, v in health_checks.items() if k.endswith("-pg")}
+    wait_for_service_health_checks(pg_checks)
+    for collection in sorted(config.collections.keys()):
+        docker.exec_(f'snoop-{collection}-pg', 'sh', '/local/db-in-vault.sh')
```
gabriel-v: `set-password.sh`?

liquid_node/commands.py
```diff
@@ -217,6 +219,13 @@ def start(job, hcl):
         job_checks = start(job, hcl)
         health_checks.update(job_checks)
 
+    # Wait for database health checks of all collections and hoover:
+    pg_checks = {k: v for k, v in health_checks.items() if k.endswith("-pg")}
```
mgax: This feels dangerous, we've used the `-pg` convention so far, and will probably do so in the future, but I'd rather flag those tasks explicitly. Maybe add a `pg_tasks` list in [liquid_node.jobs.hoover.Hoover](https://github.com/liquidinvestigations/node/blob/90afe87bc65ef4a298b409a65b9d67cb466d60e4/liquid_node/jobs/hoover.py#L4)?

templates/collection.nomad
MattesR: Yeah you're right about that and I have to admit that I'm not always consistent with it. I'll change that.

templates/collection.nomad
```diff
@@ -92,9 +92,39 @@ job "collection-${name}" {
...
+      }
+      template {
+        data = <<-EOF
+        #!/bin/sh
+        set -ex
+        pwd
+        date
+        if grep -Fq "$host all all all trust" $PGDATA/pg_hba.conf
+        then
+          psql -U snoop -c "ALTER USER snoop password '$POSTGRES_PASSWORD'"
```
mgax: This, combined with `set -x`, will write the db password in nomad logs. You can wrap this line in parantheses and `set +x` will only apply within:
```shell
(
set +x
psql -U snoop -c "ALTER USER snoop password '$POSTGRES_PASSWORD'"
)
```

templates/collection.nomad
```diff
@@ -92,9 +92,39 @@ job "collection-${name}" {
...
+          sed -i '$d' $PGDATA/pg_hba.conf
+          sed -i '$d' $PGDATA/pg_hba.conf
```
mgax: What is the goal here with `sed`? And why twice?

templates/collection.nomad
```diff
@@ -92,9 +92,39 @@ job "collection-${name}" {
...
+          {
+            echo
+            echo "host all all all md5"
```
mgax: Do we have any other option than `md5`? Maybe `sha1`? Because MD5 has been shown to be weak for a while now - https://en.wikipedia.org/wiki/MD5#Security

templates/hoover.nomad
```diff
@@ -63,9 +63,39 @@ job "hoover" {
...
+          {{- with secret "liquid/hoover/search.postgres" }}
+            POSTGRES_PASSWORD = {{.Data.secret_key}}
+          {{- end }}
+        EOF
+        destination = "local/postgres.env"
+        env = true
+      }
+      template {
+        data = <<-EOF
+        #!/bin/sh
```
mgax: This looks duplicate, you can extract it as a Jinja macro.

templates/collection.nomad
MattesR: Yeah that was my last missing piece, I wanted to ask you guys about that. The password will also be logged during the startup code, I think.
```
        if [ -z "$SNOOP_DB" ] \
                || [ -z "$SNOOP_TIKA_URL" ] \
                || [ -z "$SNOOP_ES_URL" ] \
                || [ -z "$SNOOP_AMQP_URL" ]; then
```
I'll change it there as well, okay?

templates/collection.nomad
MattesR: it deletes the last two lines of the `pg_hba.conf`. The follwing lines replace them.

templates/collection.nomad
MattesR: It's the default of postgres, and what the container will do in its entrypoint if the password is set during the first startup, that's why I used it. I'll check whether we can change that.

templates/collection.nomad
MattesR: The nomad people use `| toJSON`for variables which may contain special characters. Also, they don't put quotes around their example:
`DB_PASSWD={{ .Data.data.DB_PASSWD | toJSON }}`
Otherwise they seem to put them in quotes like this:  
`LOG_LEVEL="{{key "service/geo-api/log-verbosity"}}"`
So I can pipe the password to `toJSON`, but it wouldn't make sense to do the same with the hardcoded names and stuff. I'll make it consistent within the altered section so that environemnt variables are in quotes and there are no spaces between the name and the assignment, just like in the nomad example.

liquid_node/commands.py
MattesR: That would imply doing what @gabriel-v suggested and putting the db tasks into separate jobs?! I think that it would clutter the interface. We already have 2 jobs per collection and we have to click what feels like 20 times to see logs. 
For that reason, I'd rather keep the db as a task. What does you guys think? 
I could also offer:
`pg_checks = {k: v for k, v in health_checks.items() if k.endswith("-pg") and (k.startswith("snoop") or k.startswith("hoover"))}`   😀

templates/collection.nomad
MattesR: I'll just leave this here:
https://stackoverflow.com/questions/43333053/change-postgresql-password-encryption-from-md5-to-sha

gabriel-v: I would bet **zero** hassle, but we need to check that all of our Hoover tables will survive the upgrade. Can you check this out on your local machine? Just `deploy`, update all `postgresql` docker tags, `deploy` again, and see if anything is broken when browsing Hoover files.

If we do this migration now as a dependency for this pull request, please open a different pull request for upgrading everything to pg 10.

liquid_node/commands.py
gabriel-v: Yes, I think you should keep the `-pg` tasks as they are. Just don't have any `.endswith` and `.startswith` logic, just re-build a list of all your checks on the fly (the list is `config.collections`)

templates/collection.nomad
gabriel-v: `toJSON` adds `" "` because that's the JSON way of encoding a string. It makes sense to always use it when getting strings from Consul and from Vault.

Your example is just a sign that we're not the only ones with a consistency problem for things like these.

templates/collection.nomad
gabriel-v: Yeah, we want no passwords leaking into the logs.

templates/collection.nomad
MattesR: It is a little more hassle actually. The data needs to be dumped and then reimported. Postgres 10 throws an error when you try to start a 10.x container with a 9.6 db: 
` The data directory was initialized by PostgreSQL version 9.6, which is not compatible with this version 10.9 (Debian 10.9-1.pgdg90+1).`

mgax: Yeah, the upgrade process for postgres is rather involved. They have a [`pg_upgrade`](https://www.postgresql.org/docs/10/pgupgrade.html) command that needs the old postgres binary, the new one, the old data dir, and a place for the new data dir.

gabriel-v:
@MattesR please rebase your commits onto master so the tests pass.

MattesR:
> @MattesR please rebase your commits onto master so the tests pass.

@gabriel-v rebase onto master as in 'squash and merge'? Or just rebase commits in this branch?

gabriel-v:
Rebase the commits in this branch. There are concurrent changes on the master branch that are causing your tests here to fail.

mgax:
> Rebase the commits in this branch.

We're going to squash-and-merge in the end, so maybe it's easier to just merge with master, the end result is the same.

gabriel-v:
> We're going to squash-and-merge in the end, so maybe it's easier to just merge with master, the end result is the same.

Anything as long as your branch is up to date with all tests passing so we can review it.

PR #137, opened 2019-07-05
Run shared instances of stateless services, e.g. Tika

mugurrus:
Requires: https://github.com/hoover/snoop2/pull/172

Task: https://github.com/CRJI/EIC/issues/249


templates/collection.nomad
```diff
@@ -146,13 +112,16 @@ job "collection-${name}" {
...
         if [ -z "$SNOOP_DB" ] \
-                || [ -z "$SNOOP_TIKA_URL" ] \
+                || [ -z "$TIKA_URL" ] \
                 || [ -z "$SNOOP_ES_URL" ] \
                 || [ -z "$SNOOP_AMQP_URL" ]; then
           echo "incomplete configuration!"
           sleep 5
           exit 1
         fi
+        while true; do
```
gabriel-v: You're disabling the workers with this, please remove it before merging.

templates/hoover.nomad
```diff
@@ -196,6 +230,17 @@ job "hoover" {
               {{- end }}
             {{- end }}
 
+            {{- if service "tika" }}
```
gabriel-v: This `if` has no effect, since `with index . 0` on an empty sequence will render to an empty string.

templates/hoover.nomad
```diff
@@ -196,6 +230,17 @@ job "hoover" {
...
+              {{- with service "tika" }}
+                {{- with index . 0 }}
+                  location ~ ^/(.*) {
```
gabriel-v: We don't need tika proxied by our nginx here, since all Snoop instances are getting it directly from `range service tika`. We do forward elasticsearch through here, but we don't need that one either.

templates/hoover.nomad
```diff
@@ -87,6 +87,40 @@ job "hoover" {
     }
   }
 
+  group "tika" {
+    task "tika" {
+      driver = "docker"
+      config {
+        image = "logicalspark/docker-tikaserver"
```
gabriel-v: Please tag this image with the latest fixed tag from https://hub.docker.com/r/logicalspark/docker-tikaserver/tags

MattesR:
I just looked through review and saw the PR being idle for a while, so I just did some fixes and tested them.
Shall I also rebase onto master? I guess the checks will fail otherwise?

mgax:
> Shall I also rebase onto master?

You can just merge with master and push. And, while you're at it, please revert the changes to environment variable names, and maybe make a ticket to change them later, since it's true that the `SNOOP_` prefix makes no sense.

templates/hoover.nomad
```diff
@@ -87,6 +87,40 @@ job "hoover" {
...
+      }
+      resources {
+        memory = 800
+        cpu = 200
+        network {
+          port "tika" {}
+        }
+      }
+      service {
+        name = "tika"
```
mgax: I think you should update the collections template to point to the new service name for tika ;)

gabriel-v: Yep.

PR #141, opened 2019-07-15
Lower disk restrictions

gabriel-v:
- also take into account collection jobs when running `./liquid resources`
- print EphemeralDiskMB in `./liquid resources`
- use the macros to lower disk requirements for collection jobs

### DIFF
--- a/liquid_node/commands.py
+++ b/liquid_node/commands.py
@@ -138,12 +138,18 @@ def resources():
 
     def get_all_res():
         jobs = [nomad.parse(get_job(job.template)) for job in config.jobs]
+        for name, settings in config.collections.items():
+            for template in ['collection-migrate.nomad', 'collection.nomad']:
+                job = get_collection_job(name, settings, template)
+                jobs.append(nomad.parse(job))
         for spec in jobs:
             yield from nomad.get_resources(spec)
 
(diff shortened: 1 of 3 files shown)


gabriel-v:
On my machine, with CI enabled and the two default collections

on master:
```

before reducing disk quota:
```
Resource requirement totals: 
  batch CPU: 900
  batch EphemeralDiskMB: 350
  batch MemoryMB: 2100
  service CPU: 5900
  service EphemeralDiskMB: 7800
  service MemoryMB: 13630
```
after the patch:

```
Resource requirement totals: 
  batch CPU: 900
  batch EphemeralDiskMB: 350
  batch MemoryMB: 2100
  service CPU: 5900
  service EphemeralDiskMB: 5300
  service MemoryMB: 13630

```

File: liquid_node/commands.py [hunk 1]
```diff
@@ -138,12 +138,18 @@ def resources():
 
     def get_all_res():
         jobs = [nomad.parse(get_job(job.template)) for job in config.jobs]
+        for name, settings in config.collections.items():
+            for template in ['collection-migrate.nomad', 'collection.nomad']:
```
mgax: We should add these to `config.jobs` someday.

PR #142, opened 2019-07-16
Environment variable convention

MattesR:
Have all environment variables be declared as 
`KEY = "value"`
or, if it's a value, we get from nomad/consul/vault:
`SOME_KEY = {{.Data.secret_key | toJSON }}`, which adds double quotes.

Some more conventions:
Declaration of config section in templates:
` config {` instead of `config = {` . It was rarely used, but sometimes. 
```
template {
    data = <<-EOF
```
instead of
```
template {
    data = <<EOF
```
AFAICT, the dash makes it, so that the `.env` files is generated without the leading whitespaces in the template. (it also makes the highlighting in VSCode useful instead of rendering the whole block as one string). 
If there was any reason to not follow these convention which I've missed, please let me know.

The rocketchat template was not changed yet, as it uses it's own parser which doen't parse the envs correctly atm. Issue here: https://github.com/CRJI/EIC/issues/278


mgax:
> AFAICT, the dash makes it, so that the `.env` files is generated without the leading whitespaces in the template.

Huh, awesome! Is this a new feature? I remember searching the docs for something like this when we first experimented with nomad.

templates/rocketchat-migrate.nomad
```diff
@@ -14,26 +14,26 @@ job "rocketchat-migrate" {
...
         image = "${config.image('liquidinvestigations/caboose')}"
         args = ["python", "/local/rocketchat-caboose.py"]
         labels {
           liquid_task = "rocketchat-caboose"
         }
       }
       template {
-        data = <<EOF
+        data = <<-EOF
 # Auto-generated by rocketchat migrate script
```
mgax: We can indent this chunk now, correct?

MattesR:
> > AFAICT, the dash makes it, so that the `.env` files is generated without the leading whitespaces in the template.
> 
> Huh, awesome! Is this a new feature? I remember searching the docs for something like this when we first experimented with nomad.

I found that out by checking the differences in `.env` files created with `EOF` and `-EOF` . Since we were using it from time to time, I thought someone must know the difference ;) . It's not documented, at least not in any of the places I checked.

templates/rocketchat-migrate.nomad
MattesR: Did some more testing and unfortunately no. I suppose the env parser is doing something behind the scenes when the dash is set, because it doesn't work for that python script for example. Also to make the changes with `-EOF` clearer: The generated file takes the indentation of `data = <<-EOF` as reference, so any tab and space apart from there is still in the file. Following example:
```
       template {
        data = <<-EOF
          {{- range service "rocketchat-mongo" }}
            MONGO_ADDRESS = {{.Address | toJSON }}
            MONGO_PORT = {{.Port | toJSON }}
        {{- end }}
        EOF
        destination = "local/liquid.env"
        env = true
      }
```
results in:
```

    MONGO_ADDRESS = "10.66.60.1"
    MONGO_PORT = 20615
```
we have 2 space indentation to the service check + 2 inside the service check.
with `<<EOF`, this will be generated:
```

            MONGO_ADDRESS = "10.66.60.1"
            MONGO_PORT = 30816
        
```
The last line has 8 spaces, just as the line with `EOF` on.
```
      template {
        data = <<-EOF
        {{- range service "rocketchat-mongo" }}
        MONGO_ADDRESS = {{.Address | toJSON }}
        MONGO_PORT = {{.Port | toJSON }}
        {{- end }}
        EOF
        destination = "local/liquid.env"
        env = true
      }
```
will generate an env file with no spaces, omitting `-` will result in an env file with (you guessed it) 8 spaces.

mgax: Ok, we can live with same-level indentation. So let's indent this block to be even with `data = <<-EOF`.

gabriel-v:
> AFAICT, the dash makes it, so that the .env files is generated without the leading whitespaces in the template.

I found it by accident through the habit of writing bash all day.
The only nuisance with it is that it only cuts whitespace up to the `data =` level, never more. The bash `-EOF` cuts all leading it can, but this stops at those 12 or so characters.

MattesR:
> > AFAICT, the dash makes it, so that the .env files is generated without the leading whitespaces in the template.
> 
> I found it by accident through the habit of writing bash all day.
> The only nuisance with it is that it only cuts whitespace up to the `data =` level, never more. The bash `-EOF` cuts all leading it can, but this stops at those 12 or so characters.

yup. But only in `.env` files apparently. See comment above about indentation of the python script in `rocketchat-migrate.nomad`. Or do you have any idea why `-EOF` doesn't seem to fix it in the python script @gabriel-v

gabriel-v:
Yes: if you do this:
```
      data = <<-EOF
      {% include python.code.py %}
      EOF
```

Then what you get is:
```
      data = <<-EOF
      import something
import something_else
def main():
 ....
      EOF
```

And so it won't have any space to de-indent, since the minimum indentation is 0,

gabriel-v:
So for the time being, just leave it ugly as my first example, with the `{% include %}` not indented at all. Or we could use Jinja tags `{%- include code.py %}` to remove the whitespace - https://stackoverflow.com/a/35777386

PR #148, opened 2019-07-17
don't launch rabbitmq if workers=0

gabriel-v:
This is tricky. On `initcollection` we're running the dispatcher from `snoop-api` so we either need the rabbitmq to be up for that, or we somehow omit running the dispatcher.

I feel like the `snoop manage.py initcollection` command is too smart for us; we should split it into 4 commands and run those one by one from node's `initcollection()`.

@mgax ideas?


MattesR:
We already have `${workers}` in the template. We could just map it into the `api` as an environment variable and check for it in `initcollection`

PR #154, opened 2019-07-19
Run migrations as part of the main service

mgax:
Delete the collection migrate job and move the command to the snoop web task.


templates/collection.nomad
```diff
@@ -236,6 +236,9 @@ job "collection-${name}" {
           sleep 5
           exit 1
         fi
+        time ./manage.py migrate --noinput
```
gabriel-v: Here's your problem:
```
/local/startup.sh: 18: /local/startup.sh: time: not found
```

gabriel-v: time is a package: https://packages.ubuntu.com/bionic/time

templates/collection.nomad
```diff
@@ -236,6 +236,10 @@ job "collection-${name}" {
           sleep 5
           exit 1
         fi
+        date
+        ./manage.py migrate --noinput
+        ./manage.py healthcheck
```
gabriel-v: Since we're running it here, we should also remove it from the job: https://github.com/liquidinvestigations/node/blob/b91d76371b28f1a2176d386d39bbeaa4888b0fe2/templates/collection.nomad#L303

PR #158, opened 2019-07-23
Make liquid script a wrapper for Pipenv

### LINKED ISSUE #136: Remove `liquid` shell script, just use pipenv
Remove the `./liquid` script and invite users to manage their own virtual envs using pipenv.

Add the required steps to the readme file. Probably something like this:

```bash
sudo apt update
sudo apt install python3-venv python3-pip git curl unzip
pip3 install pipenv
pipenv install
pipenv shell
```

raduklb:
Deleted the liquid script and added instructions for helping the user configure its own virtual environment using Pipenv.

Closes #136


Readme.md
```diff
@@ -121,7 +131,7 @@ root_token = s.Cmro41vNI4wIndgrPqzlqOKY
 Then deploy to the cluster:
 
 ```shell
-./liquid deploy
+./liquid.py deploy
```
mgax: Do we really want to drop the `./liquid $something` scheme and add `.py` everywhere? We'll have to update all our instructions. At least let's keep it around for a while as "deprecated" and have it print a warning. cc @gabriel-v

raduklb: So, what should I do? @gabriel-v @mgax 
- make `./liquid` a wrapper
- consider `./liquid` deprecated and display a warning
- leave it as it currently is

mgax: Make `./liquid` a wrapper for `pipenv run python liquid.py`. Which means it won't auto-install packages the first time, like it used to, but we can document that initial `pipenv install` step.

liquid
```diff
@@ -1,19 +1,4 @@
 #!/bin/bash
 set -e
 
-# cd to file
-cd "$(dirname ${BASH_SOURCE[0]})"
```
mgax: Please keep this line, since you're invoking `./liquid.py` assuming CWD is the project folder.

vagrant/provision-liquid.sh
```diff
@@ -15,6 +15,9 @@ fi
 
 cp examples/liquid.ini .
 
+sudo apt install python3-venv python3-pip
```
mgax: Please use `apt-get` instead of `apt`. Also, add `-qy` as arguments, so it won't prompt the user for confirmation.

Readme.md
```diff
@@ -88,6 +88,16 @@ where `[liquid_domain]` is the value of `liquid.domain` from the `liquid.ini` fi
 
 
 ## Configuration
+### Use Pipenv
+```shell
+sudo apt update
+sudo apt install python3-venv python3-pip git curl unzip
+pip3 install pipenv
+pipenv install -r requirements.txt
+pipenv shell
```
mgax: You can drop this line since we're keeping the `./liquid` helper script. Don't forget to adjust the next paragraph so it doesn't lie :)

liquid
```diff
@@ -1,19 +1,6 @@
...
-
-# create venv if needed
-if [ ! -d ./venv ]; then
-  python3 -m venv venv
-  ./venv/bin/pip install -r requirements.txt
-  shasum requirements.txt > venv/requirements.shasum
-fi
-
-exec ./venv/bin/python liquid.py "$@"
+pipenv run ./liquid.py $@
```
mgax: Please write that last bit as `"$@"`, it quotes the arguments, in case they contain space.

vagrant/provision-liquid.sh
```diff
@@ -16,6 +16,10 @@ fi
 cp examples/liquid.ini .
 cat vagrant/liquid-collections.ini >> liquid.ini
 
+sudo -H apt-get install -qy python3-venv python3-pip
+sudo -H pip3 install pipenv
+pipenv install -r requirements.txt
+chmod 777 ./liquid
```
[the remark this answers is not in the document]
mgax: Doesn't the `liquid` file already have the right permissions?

vagrant/provision-liquid.sh
```diff
@@ -16,6 +16,10 @@ fi
 cp examples/liquid.ini .
 cat vagrant/liquid-collections.ini >> liquid.ini
 
+sudo -H apt-get install -qy python3-venv python3-pip
```
mgax: Do you really need `sudo -H`? plain `sudo` should be enough.

Readme.md
```diff
@@ -88,6 +88,15 @@ where `[liquid_domain]` is the value of `liquid.domain` from the `liquid.ini` fi
 
 
 ## Configuration
+### Use Pipenv
+```shell
+sudo apt update
+sudo apt-get install python3-venv python3-pip git curl unzip
+pip3 install pipenv
+pipenv install -r requirements.txt
```
mgax: The typical way to use `pipenv` is with a `Pipfile` instead of a `requirements.txt` file. Please read their documentation.

raduklb:
Should I remove the `requirements.txt` and `requirements.in` files since we switched to using `Pipfile` and `Pipfile.lock`?

PR #159, opened 2019-07-24
Allow customisation of the apps started on deploy

mugurrus:
Issue https://github.com/CRJI/EIC/issues/271

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -278,3 +278,18 @@ In order to remove a collection, take the following steps:
 #### Removing dead jobs from nomad
 In order to remove dead jobs from nomad run the following command:
 `./liquid nomadgc`.
+
+#### Enabling/disabling applications
+Applications can be enabled/disabled on deploy by setting them `on` or `off`
+in the `apps` section:
+```ini
+[apps]
+nextcloud = off
+```
+
+By default, all applications are started, but this default can also be changed
+in the `deploy` section:
+```ini
+[deploy]
+default_app_status = off
+```
--- a/Readme.md
+++ b/Readme.md
@@ -275,6 +275,13 @@ In order to remove a collection, take the following steps:
 2. Run `./liquid collectionsgc`
 3. Run `./liquid purge`
 
+#### Stopping jobs that should not be running in the current deploy configuration
+This command will stop all jobs from collections that are no longer in the `liquid.ini`
+file and jobs from applications that were disabled.
+```bash
+./liquid gc
+```
+
 #### Removing dead jobs from nomad
 In order to remove dead jobs from nomad run the following command:
 `./liquid nomadgc`.
@@ -290,6 +297,6 @@ nextcloud = off
 By default, all applications are started, but this default can also be changed
 in the `deploy` section:
 ```ini
-[deploy]
+[apps]
 default_app_status = off
 ```
--- a/liquid_node/commands.py
+++ b/liquid_node/commands.py
@@ -159,7 +159,7 @@ def resources():
     """Get memory and CPU usage for the deployment"""
 
     def get_all_res():
-        jobs = [nomad.parse(get_job(job.template)) for job in config.jobs]
+        jobs = [nomad.parse(get_job(job.template)) for job in config.enabled_jobs]
         for name, settings in config.collections.items():
             job = get_collection_job(name, settings, 'collection.nomad')
             jobs.append(nomad.parse(job))
@@ -188,6 +188,8 @@ def check_system_config():
     This checks if elasticsearch will accept our
     vm.max_map_count kernel parameter value.
     """
+    if os.uname().sysname == 'Darwin':
+        return
 
     assert int(run("sysctl -n vm.max_map_count")) >= 262144, \
         'the "vm.max_map_count" kernel parameter is too low, check readme'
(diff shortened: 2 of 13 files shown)


Readme.md
```diff
@@ -253,3 +253,18 @@ template = local/foo.nomad
...
+```ini
+[apps]
+nextcloud = off
+```
+
+By default, all applications are started, but this default can also be changed
+in the `deploy` section:
+```ini
+[deploy]
+default_app_status = off
```
gabriel-v: If we're going to have this override, let's at least put it under `[apps]` right before all those other settings it switches.

liquid_node/commands.py
```diff
@@ -281,6 +282,15 @@ def start(job, hcl):
         tokens = json.loads(run(docker_exec_cmd, shell=False))
         vault.set(app['vault_path'], tokens)
 
+    # check if there are jobs to stop
```
gabriel-v: The `deploy` command shuoldn't stop anything, we have separate commands for stopping things (gc).

liquid_node/configuration.py
```diff
@@ -159,6 +147,25 @@ def load_job(self, name, job_config):
...
+            app_jobs = list(import_string(f'liquid_node.jobs.{core_app}.jobs'))
+            all_jobs.extend(app_jobs)
+            if self.is_app_enabled(core_app):
+                jobs_to_start.extend(app_jobs)
+
+        return all_jobs, jobs_to_start
+
+    def is_app_enabled(self, job_name):
+        return job_name in Configuration._allways_on_apps or \
+            strtobool(self.ini.get('apps', job_name, fallback=self.default_app_status))
```
gabriel-v: `self.ini.getboolean`

liquid_node/configuration.py
```diff
@@ -1,43 +1,31 @@
...
-from liquid_node.jobs import rocketchat
-from liquid_node.jobs import nextcloud
-from liquid_node.jobs import ci
+from .util import import_string
+from liquid_node.jobs import ci, Job
 
 
 class Configuration:
+    _core_apps = ('liquid', 'hoover', 'dokuwiki', 'rocketchat', 'nextcloud',)
+    _allways_on_apps = ('liquid', 'hoover',)
```
gabriel-v: Typo: `always`.

I think we should name the first list `ALL_APPS` and the second one `CORE_APPS`, both in the global namespace. The `core` is something we can't do without, so a comment `# These apps can't be turned off.` should be enough to clarify this.

I guess we should be able to disable running `hoover` too, so please open an issue to track that.

liquid_node/configuration.py
```diff
@@ -159,6 +147,25 @@ def load_job(self, name, job_config):
...
+        for core_app in Configuration._core_apps:
+            if core_app in Configuration._allways_on_apps and self.ini.has_option('apps', core_app):
+                raise RuntimeError(f'The app "{core_app}" is always on.')
+
+            app_jobs = list(import_string(f'liquid_node.jobs.{core_app}.jobs'))
+            all_jobs.extend(app_jobs)
+            if self.is_app_enabled(core_app):
+                jobs_to_start.extend(app_jobs)
+
+        return all_jobs, jobs_to_start
```
gabriel-v: Since we shouldn't stop jobs when running deploy, you can remove `all_jobs` from here and only return the `jobs_to_start` list.

liquid_node/jobs/__init__.py
```diff
@@ -126,6 +128,25 @@ def get_job(hcl_path, substitutions={}):
     return output
 
 
+def wait_for_stopped_jobs(stopped_jobs):
```
gabriel-v: The `gc` command we already have should stop any jobs the user disabled in the config.

liquid_node/configuration.py
```diff
@@ -159,6 +147,25 @@ def load_job(self, name, job_config):
...
+
+    def is_app_enabled(self, job_name):
```
gabriel-v: `app_name`, not `job_name`

liquid_node/configuration.py
```diff
@@ -159,6 +147,25 @@ def load_job(self, name, job_config):
...
 
+    def _jobs(self):
+        all_jobs = []
+        jobs_to_start = []
+
+        for core_app in Configuration._core_apps:
+            if core_app in Configuration._allways_on_apps and self.ini.has_option('apps', core_app):
+                raise RuntimeError(f'The app "{core_app}" is always on.')
+
+            app_jobs = list(import_string(f'liquid_node.jobs.{core_app}.jobs'))
```
gabriel-v: Instead of changing the import dance here, I believe it would be cleaner to set a property `app` on all `Job` classes,  so you can keep the `self.jobs` list as it was and build two more lists in the constructor for this class: `self.enabled_jobs = [job for job in self.jobs if self.is_app_enabled(job.app)]` and `self.disabled_jobs = [... if not self.is_app_enabled(job.app)]`. The enabled jobs get run on `deploy`, the disabled ones get stopped on `gc`.

liquid_node/commands.py
mugurrus: I disagree, the deploy command should bring the system to the state that corresponds to the configuration from the ini file. If after deploy, the system state does not correspond to the configuration from the ini file then this is confusing.

mgax: In principle you are right, but deploy should never ever stop anything that it did not start itself. That's a tricky problem to solve and is not worth our time right now.

mugurrus: All jobs are started by deploy, there is no other command that starts jobs

mugurrus: Also, there's nothing tricky about stopping some jobs for the apps that were disabled

Issue #133, opened 2019-07-04
Run local docker registry for VM test runners


gabriel-v:
We're spending 10+ GBs of traffic per build to download all docker images. We should run a local docker registry in `templates/drone.nomad` and hook it up into the Drone workers like we do with `VMCK_URL`. We should probably separate the HOST and PORT parts there too, since drone doesn't let you put `:` in the variables.

gabriel-v:
done in ansible configs

PR #161, opened 2019-07-25
Docker local registry

Jokeswar:
Issue #133

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -205,6 +205,13 @@ The `method` is optional. The default value is `link`, while the possible values
 
 When using the `copy` method, the import will take longer.
 
+### Starting a local registry
+
+`templates/drone.nomad` should be running on your Nomad cluster. It provides a local registry
+mirror on port 6665, but dockerd is not using it yet. To enabled it just run `{root}/templates/registry-setup.sh`.
+This will let docker use the local mirror first then it will try the official one if the
+requested image is not present.
+
 ### Debugging
 
 Set the debug flag in `liquid.ini`:
--- a/templates/drone.nomad
+++ b/templates/drone.nomad
@@ -399,4 +399,58 @@ job "drone" {
       }
     }
   }
+
+  group "registry" {
+    ${ group_disk() }
+    task "registry" {
+      ${ task_logs() }
+      driver = "docker"
+      config {
+        image = "registry:2"
+        volumes = [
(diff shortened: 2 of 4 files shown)


mgax:
How are we going to solve the chicken-and-egg problem of bootstrapping the registry? It must be started with dockerd connected to the official docker hub, then dockerd needs to be configured to use the mirror instead. We need to document these steps, or automate them. And I suspect it's easier to automate them if we keep the registry outside of nomad.

gabriel-v:
Yes, this PR doesn't contain instructions on jacking up the host `dockerd` to use `localhost:6665` as the first mirror and the official registry as the second mirror.
There's no "chicken and egg" problem: if the mirror is down, the daemon will use the next available mirror (the official one). When it eventually does become available, the daemon will use it instead.

Jokeswar:
I added a README section that explains how to enable the local registry. It might just be fine if we leave Nomad do the start-up phase for the local mirror. I include a script `templates/registry-setup.sh` that gives dockerd the necessary config file and restarts the daemon so it can use the registry. The only way the script runs is if the user executes it. Is there any way to automate that?

gabriel-v:
You could use a [raw exec](https://www.nomadproject.io/docs/drivers/raw_exec.html) batch job to run a `if ! diff daemon.json /etc/docker/deamon.json; then cp daemon.json /etc/docker && systemctl restart docker; fi`, but you'd risk the Nomad agent bugging out and dying afterwards. If we want to integrate this nicely and manage it ourselves, I'd do this in `liquidinvestigations/cluster`, since we seem to push host-level configuration over there.

Jokeswar:
I am thinking of a work around. It might be overkill. What if we keep `raw_exec` and make it run a python script. Something like `python3 setup.py`:

```python3
pid = os.fork()
if pid == 0:
    sleep(1)
    os.system('bash registry-setup.sh')
    # we could skip the lines above and do it 
    # all with python alone. aka `json` and `os`
```
I am not sure if Nomad will just follow the parent and not care about the child. If that's true my thought is that the parent will finish it's thing and get cleaned by Nomad and the child will do the config and restart part. This should prevent Nomad from bugging out. What do you think @mgax @gabriel-v ?

gabriel-v:
This is great work! We decided to leave this configurration in liquidinvestigations/cluster

PR #163, opened 2019-07-31
Nextcloud admin

MattesR:
the PR puts the admin username and passwords into the nextcloud container via env variable and updates the version of `liquid-nextcloud` to be used.
It also puts the admin password into vault.

### DIFF
--- a/templates/nextcloud-migrate.nomad
+++ b/templates/nextcloud-migrate.nomad
@@ -39,10 +39,10 @@ job "nextcloud-migrate" {
         {{- range service "nextcloud-app" }}
           NEXTCLOUD_INTERNAL_STATUS_URL = "http://{{.Address}}:{{.Port}}/status.php"
         {{- end }}
-        NEXTCLOUD_HOST = nextcloud.{{ key "liquid_domain" }}
-        NEXTCLOUD_ADMIN_USER = admin
-        {{- with secret "liquid/nextcloud/nextcloud.webadmin" }}
-          NEXTCLOUD_ADMIN_PASSWORD={{.Data.secret_key | toJSON }}
+        NEXTCLOUD_HOST = "nextcloud.{{ key "liquid_domain" }}"
+        NEXTCLOUD_ADMIN_USER = "admin"
+        {{- with secret "liquid/nextcloud/nextcloud.admin" }}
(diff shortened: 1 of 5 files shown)


templates/nextcloud.nomad
```diff
@@ -38,7 +38,11 @@ job "nextcloud" {
       template {
         data = <<-EOF
         {{- with secret "liquid/nextcloud/nextcloud.admin" }}
-          OC_PASS = {{.Data.secret_key | toJSON }}
+          OC_PASS= {{.Data.secret_key}}
```
gabriel-v: Oh, we need these here for the auto-login. Please rename this env too.

templates/nextcloud-setup.sh
```diff
@@ -39,8 +43,15 @@ elif [ "$INSTALLED" == "null" ]; then
...
-
+(
+set +x
+export OC_PASS="$UPLOADS_USER_PASS"
+php occ user:resetpassword --password-from-env uploads
+export OC_PASS="$NEXTCLOUD_ADMIN_PASSWORD"
+php occ user:resetpassword --password-from-env $NEXTCLOUD_ADMIN_USER
+)
+echo "uploads and admin password set
+"
```
MattesR: I've set the environment variable with `export` instead of using @gabriel-v s command with `sudo -u www-data` because www-data is not in the sudoers file.

gabriel-v: Yes, we already have the user switch [here].

templates/nextcloud-migrate.nomad
```diff
@@ -50,8 +52,8 @@ job "nextcloud-migrate" {
         {{- with secret "liquid/nextcloud/nextcloud.maria" }}
           MYSQL_PASSWORD = {{.Data.secret_key | toJSON }}
         {{- end }}
-        {{- with secret "liquid/nextcloud/nextcloud.admin" }}
-          OC_PASS = {{.Data.secret_key | toJSON }}
+        {{- with secret "liquid/nextcloud/nextcloud.uploads" }}
+          UPLOADS_USER_PASS = {{.Data.secret_key | toJSON }}
```
gabriel-v: Since the other one is called `NEXTCLOUD_ADMIN_PASSWORD`, could this one be `UPLOADS_USER_PASSWORD`?

PR #165, opened 2019-08-02
Tika memory from ini

MattesR:


liquid_node/jobs/__init__.py
```diff
@@ -40,6 +40,7 @@ def set_volumes_paths(substitutions={}):
     substitutions['check_timeout'] = config.check_timeout
     substitutions['consul_socket'] = os.path.realpath(config.consul_socket)
     substitutions['consul_url'] = config.consul_url
+    substitutions['tika_memory'] = config.tika_memory
```
gabriel-v: Actually, [all these substitutions](https://github.com/liquidinvestigations/node/pull/165/files#diff-6cb1f42164f95e1cdf5803fff63be342R34-R43) need to go. We have the `config` in the template context, you can use `${config.tika_memory}` to get the value. Removing them should be done in another pull request, so just remove this line.

templates/hoover-deps.nomad
```diff
@@ -110,7 +110,7 @@ job "hoover-deps" {
         }
       }
       resources {
-        memory = 800
+        memory = ${tika_memory}
```
gabriel-v: see above, `${config.tika_memory}`

liquid_node/jobs/__init__.py
MattesR: Yes I agree. I also thought that the whole block was pretty messy to begin with.

PR #174, opened 2019-08-19
Update vmck

Jokeswar:
Update to the newest vmck version. It should resolve our issues with `vagrant up` from vmck/vagrant-vmck#13

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -8,7 +8,7 @@ steps:
   - pip install flake8 > /dev/null
   - flake8
 - name: deploy to qemu
-  image: vmck/vagrant-vmck:0.1.0
+  image: vmck/vagrant-vmck:0.2.0
   commands:
   - export VMCK_URL=http://$VMCK_IP:$VMCK_PORT
   - echo $VMCK_URL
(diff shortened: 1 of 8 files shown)


.drone.yml
```diff
@@ -8,7 +8,7 @@ steps:
   - pip install flake8 > /dev/null
   - flake8
 - name: deploy to qemu
-  image: vmck/vagrant-vmck:continuous-integration
+  image: vmck/vagrant-vmck:0.1.0
```
mgax: Why not `0.2.0`?

versions.ini
```diff
@@ -6,4 +6,4 @@ hoover-ui = liquidinvestigations/hoover-ui:0.2.0
 hoover-snoop2 = liquidinvestigations/hoover-snoop2:0.2.2
 liquid-nextcloud = liquidinvestigations/liquid-nextcloud:0.2.2
 # CI only
-vmck = vmck/vmck:0.3.0
+vmck = vmck/vmck:0.4.0
```
gabriel-v: remove this line, as PR #176 removed the need for this

PR #175, opened 2019-08-19
Importcollection and Exportcollection

MattesR:
The PR adds an importcollection and exportcollection command to liquid and explains how to use them in the Readme.
It also tweaks the docker commands slightly, so that `dockerexec` can handle flags just like `docker exec` and so that one can pipe files into and out of the container (needed for import and export).
`initcollection` checks whether an index is stored at `<liquid_volumes/collections/<collection_name>/index/<collection_name>-index.tgz` and if so, imports the index instead of initializing the collection.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -222,6 +222,9 @@ When using the `copy` method, the import will take longer.
 
 ### Importing and export single collections
 
+_**note:**_ Since directories might have been created by the docker daemon instead of a user,
+root privileges might be necessary.
+
 #### Exporting
 Liquid has to be deployed in order to export collections. When deployed, run
 ```
--- a/Readme.md
+++ b/Readme.md
@@ -223,7 +223,7 @@ When using the `copy` method, the import will take longer.
 ### Importing and export single collections
 
 #### Exporting
-Liquid has to be deployed in order to export collections. When deployed, run 
+Liquid has to be deployed in order to export collections. When deployed, run
 ```
 ./liquid exportcollection <collection_name>
 ```
@@ -236,10 +236,10 @@ you need paths to the following data in order to import a collection:
 - index: a `.tgz` archive containing the index
 run the following command:
 ```
-./liquid importcollection <collection_name> <database_path> <blobs_path> <index_path> [method] 
+./liquid importcollection <collection_name> <database_path> <blobs_path> <index_path> [method]
 ```
 You can also specify whether data is moved, copied or linked by specifying the method.
-As of now, this is a positional argument, so put `method=copy/link/move` at the end of your command 
+As of now, this is a positional argument, so put `method=copy/link/move` at the end of your command
 
 ### Debugging
 
--- a/Readme.md
+++ b/Readme.md
@@ -237,9 +237,9 @@ you need paths to the following data in order to import a collection:
 run the following command:
 ```
 ./liquid importcollection <collection_name> <database_path> <blobs_path> <index_path> [method] 
-
+```
 You can also specify whether data is moved, copied or linked by specifying the method.
-As of now, this is a positional argument, so put `method='copy/link/move` at the end of your command 
+As of now, this is a positional argument, so put `method=copy/link/move` at the end of your command 
 
 ### Debugging
 
--- a/Readme.md
+++ b/Readme.md
@@ -220,6 +220,27 @@ The `method` is optional. The default value is `link`, while the possible values
 
 When using the `copy` method, the import will take longer.
 
+### Importing and export single collections
+
+#### Exporting
+Liquid has to be deployed in order to export collections. When deployed, run 
+```
+./liquid exportcollection <collection_name>
+```
+The exported collection data is stored in `<collections_directory>/exported/<collection_name`
+
+#### Importing
+you need paths to the following data in order to import a collection:
+- database: the folder must contain a subfolder called data, containing the database
+- blobs: the folder contains the blobs
+- index: a `.tgz` archive containing the index
+run the following command:
+```
+./liquid importcollection <collection_name> <database_path> <blobs_path> <index_path> [method] 
+
+You can also specify whether data is moved, copied or linked by specifying the method.
+As of now, this is a positional argument, so put `method='copy/link/move` at the end of your command 
+
 ### Debugging
 
 Set the debug flag in `liquid.ini`:
--- a/Readme.md
+++ b/Readme.md
@@ -230,7 +230,7 @@ Liquid has to be deployed in order to export collections. When deployed, run
 ```
 ./liquid exportcollection <collection_name>
 ```
-The exported collection data is stored in `<collections_directory>/exported/<collection_name`
+The exported collection data is stored in `volumes/exports/<collection_name>`
 
 #### Importing
 you need paths to the following data in order to import a collection:
--- a/liquid_node/commands.py
+++ b/liquid_node/commands.py
@@ -6,7 +6,6 @@ import os
 import base64
 import json
 import shutil
-import tarfile
 import gzip
 
 from liquid_node.collections import push_collections_titles
@@ -552,31 +551,24 @@ def exportcollection(name):
     if export_path.exists():
         if confirm(f'collection {name} already has exported files in {export_path}, overwrite?'):
             shutil.rmtree(export_path)
-            export_path.mkdir(parents=True)
         else:
             log.info(f'exiting and not overwriting data for {name}')
             exit()
     else:
-        export_path.mkdir(parents=True)
+        (export_path).mkdir(parents=True)
     collection_volumes = Path(config.liquid_volumes) / 'collections' / name
-
-    # dumping the database
-    with gzip.open(export_path / 'pg_dump.tgz', "w") as pg_dump:
-        docker.exec_(
-            f'snoop-{name}-pg',
-            'pg_dump', '-Ox', '-U', 'snoop', 'snoop',
-            stdout=pg_dump, interactive=True
-        )
-    log.info(f'dumped database to {export_path}')
+    # copy the pg dir
+    database = collection_volumes / 'pg'
+    pg_src = database
+    pg_dst = export_path / 'pg'
+    import_dir(pg_src, pg_dst, method='copy')
 
     # copy the blobs dir
     blobs = collection_volumes / 'blobs'
-    blobs_tar = tarfile.open(export_path / "blobs.tgz", "w:gz")
-    blobs_tar.add(blobs, arcname='blobs')
-    blobs_tar.close()
-    log.info(f'dumped blobs to {export_path}')
+    blob_src = blobs
+    blob_dst = export_path / 'blobs'
+    import_dir(blob_src, blob_dst, method='copy')
 
-    # copy the index
     with gzip.open(export_path / f'{name}-index.tgz', "w") as index_file:
         docker.exec_(
             f'snoop-{name}-api',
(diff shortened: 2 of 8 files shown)


liquid_node/docker.py
```diff
@@ -10,33 +10,47 @@ def containers(self, labels=[]):
         out = run(f'docker ps -q {label_args}')
         return out.split()
 
-    def exec_command(self, name, *args, tty=False):
+    def exec_command(self, *args, tty=False):
```
gabriel-v: How about we keep the `name` argument here, and also pass `stdin` and `tty` as boolean arguments? If they are set, you add `-i` and `-t` respectively, instead of `-it` as it is now. I'd avoid parsing the command line as you're doing at line 26 here.

liquid_node/commands.py
```diff
@@ -450,6 +474,110 @@ def deletecollection(name):
...
+    blobs = Path(blobs).resolve(strict=True)
+    blob_src = blobs
+    blob_dst = collection_path / 'blobs'
+    import_dir(blob_src, blob_dst, method=method)
+
+    # copy or link the index
+    index = Path(index).resolve(strict=True)
+    index_path = collection_path / 'index'
+    index_path.mkdir(parents=True)
+    if method == 'copy':
```
gabriel-v: Please add a new function `import_from_docker.import_file` and use it here, instead of adding the three-way if here.

Readme.md
```diff
@@ -220,6 +220,30 @@ The `method` is optional. The default value is `link`, while the possible values
...
+The exported collection data is stored in `<collections_directory>/exported/<collection_name`
+
+#### Importing
+you need paths to the following data in order to import a collection:
+- database: the folder must contain a subfolder called data, containing the database
+- blobs: the folder contains the blobs
+- index: a `.tgz` archive containing the index
+run the following command:
+```
+./liquid importcollection <collection_name> <database_path> <blobs_path> <index_path> [method]
```
gabriel-v: Could we simplify this to `importcollection <name> <path> [<method>]` and expect the data at hard-coded directory names under `<path>`, having the same structure as the `export` result?

The idea would be to have these two commands just work, one after the other:
- `./liquid exportcollection testdata`
- `./liquid importcollection testdata-clone collections/exported/testdata`

MattesR: Yeah I agree, like this it's clunky. It was to be replaced by the execution of my masterplan to archive it all in one password protected file, but since that is scrapped, I should change that.

Readme.md
```diff
@@ -220,6 +220,30 @@ The `method` is optional. The default value is `link`, while the possible values
...
+
+_**note:**_ Since directories might have been created by the docker daemon instead of a user,
+root privileges might be necessary.
+
+#### Exporting
+Liquid has to be deployed in order to export collections. When deployed, run
+```
+./liquid exportcollection <collection_name>
+```
+The exported collection data is stored in `<collections_directory>/exported/<collection_name`
```
mgax: I think we should use a folder outside of the collections directory for exporting. Maybe the user mounts that directory read-only. Or maybe they have a collection called `exported` - journalists investigating cross-border trade perhaps? :) So how about we create a new folder, `exports`, in the liquid volumes folder (e.g. `/opt/node/volumes/exports`, and  write stuff in there?

Also, missed a closing `>` here ;)

liquid_node/commands.py
```diff
@@ -450,6 +474,110 @@ def deletecollection(name):
...
+            log.info(f'exiting and not overwriting data for {name}')
+            exit()
+    else:
+        redeploy = False
+        (collection_path).mkdir(parents=True)
+
+    # copy the pg dir
+    database = Path(database).resolve(strict=True)
+    if not (database / 'data' / 'PG_VERSION').exists():
+        raise RuntimeError("database is not a valid Postgres database")
```
mgax: The normal way of exporting and importing a postgres database is to use  `pg_dump -Ox $dbname` which generates a text file. You can pipe the output through `gzip  -1` for a low-cpu effort to compress the data. I don't think we should be passing on-disk postgres database files around.

liquid_node/commands.py
```diff
@@ -450,6 +474,110 @@ def deletecollection(name):
...
+    """Export a collection from node. Liquid has to be running.
+
+    :param name: name of the collection you want to export.
+
+    """
+    if name not in config.collections:
+        raise RuntimeError(f'Collection {name} does not exist in the liquid.ini file.')
+    export_path = Path(config.liquid_collections) / 'exported' / name
+    if export_path.exists():
+        if confirm(f'collection {name} already has exported files in {export_path}, overwrite?'):
```
mgax: How about we just print an error and bail? It's easy to `sudo rm -rf the/export/path` and try again, and we reduce some complexity.

liquid_node/commands.py
```diff
@@ -450,6 +474,110 @@ def deletecollection(name):
...
+    else:
+        (export_path).mkdir(parents=True)
+    # copy the pg dir
+    database = Path(config.liquid_volumes) / 'collections' / name / 'pg'
+    pg_src = database
+    pg_dst = export_path / 'pg'
+    import_dir(pg_src, pg_dst, method='copy')
+
+    # copy the blobs dir
+    blobs = Path(config.liquid_volumes) / 'collections' / name / 'blobs'
```
mgax: You're duplicating the path a lot. Write the common part (`Path(config.liquid_volumes) / 'collections' / name`) to a variable (say `collection_volumes`) so it's easy to mentally figure out that live docker volumes are in there.

liquid_node/commands.py
```diff
@@ -450,6 +474,110 @@ def deletecollection(name):
...
+    blob_src = blobs
+    blob_dst = export_path / 'blobs'
+    import_dir(blob_src, blob_dst, method='copy')
+
+    log.info(f'found index for {name}, importing index instead of creating it')
```
mgax: This log message looks weird, are you sure it's right?

liquid_node/docker.py
```diff
@@ -10,33 +10,37 @@ def containers(self, labels=[]):
...
 
-        docker_exec_cmd = ['docker', 'exec']
         if tty:
-            docker_exec_cmd += ['-it']
-        docker_exec_cmd += [container_id] + list(args or (['bash'] if tty else []))
-
+            docker_exec_cmd += ['-t']
+        if interactive:
+            docker_exec_cmd += ['-i']
+        docker_exec_cmd += [container_id] + list(args or (['bash'] if (tty and interactive) else []))
```
mgax: Maybe it's time to drop this `bash` default thing. Docker invokes the default entrypoint anyway.

liquid_node/commands.py
```diff
@@ -450,6 +474,110 @@ def deletecollection(name):
...
+    # copy the pg dir
+    database = Path(database).resolve(strict=True)
+    if not (database / 'data' / 'PG_VERSION').exists():
+        raise RuntimeError("database is not a valid Postgres database")
+    pg_src = database
+    pg_dst = collection_path / 'pg'
+    import_dir(pg_src, pg_dst, method=method)
+
+    # copy the blobs dir
+    blobs = Path(blobs).resolve(strict=True)
```
mgax: `blobs` is already a `Path` object :) And do you really need `strict`? I had to look up the documentation to see what it means, imho it just adds mental overhead when reading the code. Also I suspect that calling `resolve` is ultimately not needed because all filesystem operations will work through symlinks anyway.

liquid_node/commands.py
```diff
@@ -450,6 +474,110 @@ def deletecollection(name):
...
+    blob_src = blobs
+    blob_dst = collection_path / 'blobs'
+    import_dir(blob_src, blob_dst, method=method)
```
mgax: I think it's safer to `tar` the blob folder to a single file. For a large collection, it will consist of a huge number of little files, it's easier to manage them as a single archive. And you can pass it through `gzip -1` to make the text chunks a bit smaller, which will make a difference, since a lot of the blobs are actually intermediate processing results in JSON format.

And please leave a blank line before you start processing the index. When you look at the screen from 5 meters away you should see 3 code chunks, for blobs, index and pg. :)

Readme.md
```diff
@@ -220,6 +220,30 @@ The `method` is optional. The default value is `link`, while the possible values
...
+#### Importing
+you need paths to the following data in order to import a collection:
+- database: the folder must contain a subfolder called data, containing the database
+- blobs: the folder contains the blobs
+- index: a `.tgz` archive containing the index
+run the following command:
+```
+./liquid importcollection <collection_name> <database_path> <blobs_path> <index_path> [method]
+```
+You can also specify whether data is moved, copied or linked by specifying the method.
```
mgax: Is there a good reason to support three import/export methods? If so, we need automated tests for all of them as part of the regular test suite for each pull request, otherwise the code is actually dangerous. What if I want to copy-export a live collection and it ends up down the `move` codepath because of a seemingly innocent change?

Readme.md
MattesR: Okay I can see why that's a better place, good example ;)

liquid_node/commands.py
MattesR: I thought it would be nicer and maybe we want to confirm other stuff at some point as well. I had to import/export a lot of test collections over and over again so that felt quicker. But I don't mind deleting it.

liquid_node/commands.py
MattesR: I'm sure it's wrong ^^
The art of copy-and-paste.

Readme.md
MattesR: I've added the three option since there were already part of the fuctions from `importfromdocker`. There's (maybe) no good reason to have move, ever I agree. Honestly, since you will be the one who has to write the tests, it's your call 😁 
I'd go for just copy then?!

liquid_node/commands.py
MattesR: Oh yeah it is, that was a change that came later on, thank you. 
So my ingenious idea of using strict was that it would bail out if the part doesn't exist and therefore slims the code, because I don't have to check whether the user has actually provided a legit path. Is that unnecessary or would you have done it differently?

liquid_node/commands.py
MattesR: You mean having an archive as the input of importing, unpacking it in the process and respectively creating an archive during the export?! Will do!

liquid_node/commands.py
MattesR: This is the way it was done in `docker-setup` as well. Wasn't there a reason for doing like this?^^ 
I'll try whether exporting works with `pgdump` with the testdata set.

liquid_node/docker.py
MattesR: that would break `liquid shell <container>` without any argument, since docker exec needs a command.
```
DEBUG + ['docker', 'exec', '-t', '-i', '723a0878cfde']
"docker exec" requires at least 2 arguments.
See 'docker exec --help'.

Usage:  docker exec [OPTIONS] CONTAINER COMMAND [ARG...]
```
Or did you mean that everybody should just write `liquid shell <container> bash` if they want to bash into the container?

liquid_node/commands.py
MattesR: there's a new branch `pg_dump` where my progress on changing this is in. The export works but the import doesn't yet. Unfortunately I have no time to fix this so I reverted the changes.

MattesR:
SInce I'm leaving now I can't finalize that PR. Import/export is working but you must look over it for the last changes.

gabriel-v:
This is wonderful work, I'll pick this up at a later date.

PR #188, opened 2019-09-11
replace nginx with fabiolb

gabriel-v:
This should reduce our alloc count by a lot

### DIFF
--- a/templates/hoover.nomad
+++ b/templates/hoover.nomad
@@ -104,7 +104,7 @@ job "hoover" {
       upstream='hoover-search',
     ) }
 
-  group "load-balancer" {
+  group "collections-lb" {
     task "fabio" {
       driver = "docker"
       config {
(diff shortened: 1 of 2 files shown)


templates/hoover.nomad
```diff
@@ -104,104 +104,42 @@ job "hoover" {
...
-        name = "hoover-collections"
-        port = "nginx"
-        check {
-          name = "tcp"
-          initial_status = "critical"
-          type = "tcp"
-          interval = "${check_interval}"
-          timeout = "${check_timeout}"
+          port "ui" {
+          }
```
mgax: Let's give it a static port too so we don't hunt around for it.

templates/hoover.nomad
```diff
@@ -104,104 +104,42 @@ job "hoover" {
       upstream='hoover-search',
     ) }
 
-  group "collections" {
-    task "nginx" {
+  group "load-balancer" {
```
mgax: I liked the previous name, `collections`, it was more descriptive. Maybe `collections-lb` if you must?

templates/hoover.nomad
```diff
@@ -104,104 +104,42 @@ job "hoover" {
...
+        destination = "local/fabio.properties"
+        data = <<-EOH
+        registry.backend = consul
+        registry.consul.addr = ${consul_url}
+        registry.consul.checksRequired = all
+        registry.consul.tagprefix = snoop-
+        ui.addr = :9991
+        ui.color = blue
+        proxy.addr = :9990
+        EOH
```
mgax: You can indent the whole block as long as you indent `EOH` too and don't have empty lines (or the empty lines are also indented).

templates/hoover.nomad
```diff
@@ -104,104 +104,42 @@ job "hoover" {
...
+      }
+
       resources {
+        cpu    = 100
```
mgax: I thought we didn't follow this heavy whitespace alignment, but as you with.

PR #197, opened 2019-09-15
Run `promtail` to send logs to Loki

mgax:
Capture authproxy logs for now, but I think we should add `{{ promtail_task() }}` to all groups.

### DIFF
--- a/templates/_lib.hcl
+++ b/templates/_lib.hcl
@@ -85,6 +85,8 @@ ephemeral_disk {
         }
       }
     }
+
+    ${ promtail() }
   }
 {%- endmacro %}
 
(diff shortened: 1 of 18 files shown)


templates/_lib.hcl
```diff
@@ -100,3 +102,40 @@ ephemeral_disk {
     destination = "local/set_pg_password.sh"
   }
 {%- endmacro %}
+
+{% macro promtail() %}
+    task "promtail" {
+      driver = "docker"
+
+      config {
+        image = "grafana/promtail:master"
```
gabriel-v: Could we pin this with a digest?

mgax: We could, I'll add an issue because it needs testing and I don't want to delay the deployment.
https://github.com/liquidinvestigations/node/issues/198

templates/_lib.hcl
```diff
@@ -85,6 +85,8 @@ ephemeral_disk {
         }
       }
     }
+
+    ${ promtail() }
```
gabriel-v: Maybe rename this to `promtail_task` to make it clear what's being generated?

mgax: Fair enough.

PR #202, opened 2019-09-19
Integrate Codimd

gabriel-v:
Probably need to add more things from https://hackmd.io/c/codimd-documentation/%2Fs%2Fcodimd-configuration. 

[There are no theming/css options](https://github.com/hackmdio/codimd/issues/946#issuecomment-434108657) so I think we'll need to fork it and build our own in order to change the landing page and add our link to the header/footer.


mgax:
Don't forget to update the liquid-core docker tag.

PR #208, opened 2019-09-26
Fix migration script file scan

alexneamtu:
php occ files:scan doesn't work without either --all or an user-id.


mgax:
@gabriel-v you have more experience with snoop and synched folders, can you take a look pls?

gabriel-v:
This is about nextcloud finding data when something else uploads it on disk, no relationship to hoover & others.

PR #211, opened 2019-10-07
Authproxy memory

gabriel-v:
Temporary fix until we resolve https://github.com/liquidinvestigations/authproxy/issues/21

### DIFF
--- a/templates/_lib.hcl
+++ b/templates/_lib.hcl
@@ -79,8 +79,8 @@ ephemeral_disk {
           initial_status = "critical"
           type = "http"
           path = "/__auth/logout"
-          interval = "${check_interval}"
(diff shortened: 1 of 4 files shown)


mgax:
I think this pushes us over the memory limit again. Oh joy. Maybe make the parameters configurable so we don't waste memory in testing?

gabriel-v:
> I think this pushes us over the memory limit again. Oh joy. Maybe make the parameters configurable so we don't waste memory in testing?

This is temporary, so let's increase the memory for the VM.

I think we should keep increasing it until we hit 31GB, and re-think this then, maybe with some workarounds from this thread: https://github.com/hashicorp/nomad/issues/606

PR #214, opened 2019-10-07
run elasticsearch data nodes

gabriel-v:
One master node, multiple data nodes, all load-balanced by fabio.
It just works(tm)

### DIFF
--- a/templates/hoover-deps.nomad
+++ b/templates/hoover-deps.nomad
@@ -1,23 +1,11 @@
 {% from '_lib.hcl' import authproxy_group, continuous_reschedule, set_pg_password_template, promtail_task with context -%}
 
-job "hoover-deps" {
-  datacenters = ["dc1"]
-  type = "service"
-  priority = 60
-
-  group "es-master" {
-    task "es" {
-      constraint {
-        attribute = "{% raw %}${meta.liquid_volumes}{% endraw %}"
-        operator = "is_set"
-      }
-
-      driver = "docker"
+{%- macro elasticsearch_docker_config(data_dir_name) %}
       config {
         image = "docker.elastic.co/elasticsearch/elasticsearch-oss:6.8.3"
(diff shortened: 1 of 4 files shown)


vagrant/Vagrantfile
```diff
@@ -49,7 +49,7 @@ if Vagrant.has_plugin?('vagrant-env')
   config.vm.provider :vmck do |vmck, override|
     vmck.image_path = 'cluster-master.qcow2.tar.gz'
     vmck.vmck_url = ENV['VMCK_URL']
-    vmck.memory = 23500
+    vmck.memory = 25000
```
mgax: Oh boy :))

templates/hoover-deps.nomad
```diff
@@ -20,26 +20,126 @@ job "hoover-deps" {
...
+          "{% raw %}${meta.liquid_volumes}{% endraw %}/hoover/es/data-{% raw %}${NOMAD_ALLOC_INDEX}{% endraw %}:/usr/share/elasticsearch/data",
+        ]
+        port_map {
+          http = 9200
+          transport = 9300
+        }
+        labels {
+          liquid_task = "hoover-es"
+        }
+      }
```
mgax: I suggest we extract this chunk as a macro, so we don't have to change the image version in several places, and we keep the entry point hack in a single place. The es data path can be a macro argument.

templates/hoover-deps.nomad
```diff
@@ -20,26 +20,126 @@ job "hoover-deps" {
...
+          interval = "${check_interval}"
+          timeout = "${check_timeout}"
+        }
+      }
+    }
+
+    ${ promtail_task() }
+  }
+
+  {% if config.elasticsearch_data_node_count %}
```
mgax: So if `elasticsearch_data_node_count` is zero does elasticsearch still work? If so, then I'm pretty sure you can drop this `if` block, because `group { count = 0 }` is fine.

gabriel-v: We'd wait for a health check that doesn't exist, because we don't check the service count.
I think it's cleaner this way, since 0 is the default, most users won't even see this.

PR #215, opened 2019-10-08
Enable metrics collection for hoover's elasticsearch

gabriel-v:


templates/hoover-deps.nomad
```diff
@@ -51,6 +51,9 @@ job "hoover-deps" {
         transport.bind_host = "0.0.0.0"
         transport.publish_host = "{% raw %}${attr.unique.network.ip-address}{% endraw %}"
 
+        xpack.license.self_generated.type = "basic"
```
mgax: Why does this feel like doing a deal with the devil? Can we still claim we're open source? :p

gabriel-v: https://www.elastic.co/what-is/open-x-pack

All our sources are still open, yes, if that's the issue

PR #220, opened 2019-10-28
update repo urls

gabriel-v:


examples/liquid.ini
```diff
@@ -129,7 +129,7 @@ template = templates/zipkin.nomad
 
 # Add the testdata collection. Remove if not needed.
 # Be sure to clone it first:
-#     git clone https://github.com/hoover/testdata collections/testdata
+#     git clone https://github.com/liquidinvestigations/testdata collections/testdata
```
mgax: And again, `hoover-testdata`

repos/clone.sh
```diff
@@ -3,7 +3,19 @@ set -e
...
+elif [ "$1" == 'https' ]; then
+        echo "using https"
+else
+        set +x
+        echo "Usage: $0 https - clone repos with https"
+        echo "       $0 ssh   - clone repos with ssh"
+        exit 1
+fi
+
+for repo in liquidinvestigations/hoover-snoop2 liquidinvestigations/hoover-search liquidinvestigations/hoover-ui liquidinvestigations/core liquidinvestigations/authproxy liquidinvestigations/hypothesis-h; do
```
mgax: Since you're here, can you please make this list into a real list?

```bash
repos=(
  hoover-snoop2
  hoover-search
  ...
)

for repo in "${repos[@]}"; do
...
done
```

templates/hypothesis.nomad
```diff
@@ -157,7 +157,7 @@ job "hypothesis" {
 
       driver = "docker"
       config = {
-        image = "hypothesis/hypothesis"
+        image = "${config.image('hypothesis-h')}"
```
mgax: 👌

examples/liquid.ini
gabriel-v: I've already updated everything to use this name. There's no confusion, so I'll leave it like it is.

repos/clone.sh
gabriel-v: ok

PR #244, opened 2020-01-07
Demo branch

gabriel-v:
Use the newly found drone feature to implement deploying on oneself. Killing drone 1.4 doesn't kill its jobs when it restarts, so we can use this safely.


.drone.yml
```diff
@@ -32,31 +32,56 @@ steps:
...
+  - pip3 install pipenv
+  - pipenv install
+  - git --version
+  - docker --version
+  - python --version
+  - ./liquid alloc drone drone
+  # - liquid liquid.py halt
+  # - sleep 13
+  - git fetch -ap
+  - git reset --hard origin/demo
```
mgax: I'd prefer if this step failed when there is a dirty working tree. Call me paranoid or old-fashioned if you must. :)

https://remarkablemark.org/blog/2017/10/12/check-git-dirty/

PR #246, opened 2020-01-10
Configure worker limits

gabriel-v:
Process count per container + memory limit.

Requires https://github.com/liquidinvestigations/hoover-snoop2/pull/263


examples/liquid.ini
```diff
@@ -138,20 +138,23 @@ wait_green_count = 6
...
 ; sync = True
-; rabbitmq_memory_limit = 900
 
 # Add the testdata collection. Remove if not needed.
 # Be sure to clone it first:
 #     git clone https://github.com/liquidinvestigations/testdata collections/testdata
 ; [collection:testdata]
 ; workers = 1
 ; rabbitmq_memory_limit = 300
+; worker_memory_limit = 300
```
mgax: Please use the same value as the default, `400`.

PR #272, opened 2020-05-27
Hypothesis backup & restore

morten-spiegel:
The restore Process got its own [function](https://github.com/liquidinvestigations/node/compare/hypothesis-backup-restore#diff-bf5271dce6a9f8ba053979c452ab36dcR371) instead of using `restore_collection_es` because I liked it better than all the `if` statements that would have been necessary if we did it in the same function. But if you want it in the same function, just let me know.

Because [this](https://github.com/liquidinvestigations/node/compare/hypothesis-backup-restore#diff-bf5271dce6a9f8ba053979c452ab36dcR332-R338) way of resetting and closing the index wasn't possible without creating a new manage.py with the respective commands, I just [delete the old index](https://github.com/liquidinvestigations/node/compare/hypothesis-backup-restore#diff-bf5271dce6a9f8ba053979c452ab36dcR404-R407). Its working, but again, if you have any concerns let me know.

### DIFF
--- a/liquid_node/backup.py
+++ b/liquid_node/backup.py
@@ -42,7 +42,7 @@ def backup(*args):
             collection_dir = dest / "hypothesis"
             collection_dir.mkdir(parents=True, exist_ok=True)
             backup_pg(collection_dir / 'hypothesis.pg.sql.gz', 'hypothesis', 'hypothesis', 'hypothesis:pg')
-            backup_collection_es(dest / 'hypothesis' , 'hypothesis', '/_h_es')
+            backup_collection_es(dest / 'hypothesis', 'hypothesis', '/_h_es')
             backup_files(collection_dir / 'hypothesis.tgz', '/opt/rabbitmq/', [], 'hypothesis:rabbitmq')
 
     if not options.collections:
@@ -254,7 +254,7 @@ def backup_collection_es(dest, name, es_url):
             "type": "fs",
             "settings": {
                 "location": f"/es_repo/backup-{name}",
-                },
+            },
         })
         es.put(f"/_snapshot/backup-{name}/snapshot", {
             "indices": name,
@@ -367,6 +367,7 @@ def restore_collection_es(src, name):
         )
         subprocess.check_call(rm_cmd, shell=True)
 
+
 @retry()
 def restore_hypothesis_es(src, name):
     src_file = src / "es.tgz"
@@ -403,14 +404,14 @@ def restore_hypothesis_es(src, name):
 
         # delete index instead of resetting it
         old_index = es.get(f"/_cat/indices?format=json")
-        old_index_name = dele[0]["index"]
+        old_index_name = old_index[0]["index"]
         es.delete(f"/{old_index_name}")
 
         # restore snapshot
         es.post(f"/_snapshot/restore-{name}/snapshot/_restore", {
             "indices": old_name,
             "include_global_state": False,
-            #"rename_pattern": ".+",
+            # "rename_pattern": ".+",
             "rename_replacement": name,
         })
 
@@ -435,7 +436,6 @@ def restore_hypothesis_es(src, name):
             f"rm -rf /es_repo/restore-{name} "
         )
         subprocess.check_call(rm_cmd, shell=True)
-
 
 
 def backup_collection(dest, name, save_blobs=True, save_es=True, save_pg=True):
--- a/liquid_node/backup.py
+++ b/liquid_node/backup.py
@@ -320,7 +320,7 @@ def restore_es(src, name, es_url_suffix, es_alloc_id):
             subprocess.check_call(reset_cmd, shell=True)
             es.post(f"/{name}/_close")
         else:
-            # delete index instead of resetting it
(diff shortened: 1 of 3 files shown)


liquid_node/backup.py
```diff
@@ -326,24 +356,96 @@ def restore_collection_es(src, name):
...
+
         rm_cmd = (
             f"./liquid dockerexec {SNOOP_ES_ALLOC} "
             f"rm -rf /es_repo/restore-{name} "
         )
         subprocess.check_call(rm_cmd, shell=True)
 
 
+@retry()
+def restore_hypothesis_es(src, name):
```
gabriel-v, line 372: Why the huge duplication? See suggestions above for refactoring `restore_es`:
- add argument for alloc id
- add argument for es url or suffix
- keep two functions: `restore_es` and `backup_es` with the same arguments

liquid_node/backup.py
```diff
@@ -258,18 +272,26 @@ def backup_collection_es(dest, name):
             else:
                 raise RuntimeError("Something went wrong: %r" % snapshot)
         log.info(f"Snapshot done in {int(time()-t0)}s")
-
-        backup_files(tmp_file, f"/es_repo/backup-{name}", [], SNOOP_ES_ALLOC)
+        if es_url == '/_h_es':
+            backup_files(tmp_file, f"/es_repo/backup-{name}", [], HYPOTHESIS_ES_ALLOC)
```
gabriel-v, line 276: This function will depend on whatever URL scheme is chosen outside it, too complex. This should be a separate function argument: `es_alloc_id`, and the caller chooses between the two constants.

liquid_node/backup.py
```diff
@@ -258,18 +272,26 @@ def backup_collection_es(dest, name):
...
         dest_file = dest / "es.tgz"
         tmp_file.rename(dest_file)
     finally:
         es.delete(f"/_snapshot/backup-{name}/snapshot")
         es.delete(f"/_snapshot/backup-{name}")
-        rm_cmd = (
-            f"./liquid dockerexec {SNOOP_ES_ALLOC} "
-            f"rm -rf /es_repo/backup-{name} "
-        )
+        if es_url == "/_h_es":
```
gabriel-v, line 285: Same as above: what if we add another elasticsearch here? We'd need more `if` branches

liquid_node/backup.py
```diff
@@ -223,10 +237,10 @@ def is_index_available(es_client, name):
 
 
 @retry()
-def backup_collection_es(dest, name):
+def backup_collection_es(dest, name, es_url):
     tmp_file = dest / "es.tgz.tmp"
     log.info(f"Dumping collection {name} es snapshot to {tmp_file}")
-    es = JsonApi(f"http://{nomad.get_address()}:9990/_es")
+    es = JsonApi(f"http://{nomad.get_address()}:9990{es_url}")
```
gabriel-v, line 243: Best call this argument `es_url_suffix` since it's just the part after the port, not the whole URL.

liquid_node/backup.py
```diff
@@ -223,10 +237,10 @@ def is_index_available(es_client, name):
 
 
 @retry()
-def backup_collection_es(dest, name):
+def backup_collection_es(dest, name, es_url):
```
gabriel-v, line 240: I think it's best to rename this to `def backup_es(dest, name, es_url_suffix, es_alloc_id):`

liquid_node/backup.py
```diff
@@ -38,6 +38,13 @@ def backup(*args):
         if config.is_app_enabled('dokuwiki'):
             backup_files(dest / 'dokuwiki.tgz', '/bitnami/dokuwiki', [], 'dokuwiki:php')
 
+        if config.is_app_enabled('hypothesis'):
+            collection_dir = dest / "hypothesis"
+            collection_dir.mkdir(parents=True, exist_ok=True)
+            backup_pg(collection_dir / 'hypothesis.pg.sql.gz', 'hypothesis', 'hypothesis', 'hypothesis:pg')
+            backup_collection_es(dest / 'hypothesis', 'hypothesis', '/_h_es')
+            backup_files(collection_dir / 'hypothesis.tgz', '/opt/rabbitmq/', [], 'hypothesis:rabbitmq')
```
gabriel-v, line 46: These files will be corrupted, since the `rabbitmq` process might still be running. It's best to leave them out altogether and avoid the internals of how hypothesis saves its data.

If we somehow need the queue, then we'll add another save/restore pair of functions that use the rabbitmq API or CLI to dump and restore the data.

liquid_node/backup.py
```diff
@@ -38,6 +38,13 @@ def backup(*args):
         if config.is_app_enabled('dokuwiki'):
             backup_files(dest / 'dokuwiki.tgz', '/bitnami/dokuwiki', [], 'dokuwiki:php')
 
+        if config.is_app_enabled('hypothesis'):
+            collection_dir = dest / "hypothesis"
```
gabriel-v, line 42: `collection_dir` makes you think there's a hoover collection is involved, but this is not the case. Please use another name for this variable.

PR #266, opened 2020-03-25
Update Nextcloud to Nextcloud 18

morten-spiegel:
Updates the Version.ini file to Update the Nextcloud to Version 16

PR #274, opened 2020-05-28
Nextcloud 18

gabriel-v:
Update to nextcloud version 18. We're also using postgres 12. This work was started by @morten-spiegel in https://github.com/liquidinvestigations/node/pull/266

Issues with that pull request that I fixed:
- Nextcloud changed the Docker image configuration: Now `/var/www/html` is not a `volume` with pre-populated data, but empty. The image first needs to run `/entrypoint.sh` with a magic argument, and that script will `rsync` the data contents from `/usr/src/`. This is ugly, might redo the logic on our side.
- I folded together the `-migrate` and the server jobs. Now the `/local/setup.sh` script gets run in parallel with the entrypoint.
- There was a confusion with the username. It's `uploads`, not `upload`. You can see that in the docker data mount directory, it's `uploads` there. I guess this might also be present in the templates, I have to check
- Probably related to the past issue, Nextcloud changed behaviour in version 18: you can't create a user that has an existing data directory. You get greeted by the `print`ed error message: `Username is invalid because files already exist for this user`. I can't seem to grep this in the nextcloud code, I'll keep digging. This is a major problem for us, because that's how we used to use this service: with a pre-existing data directory being mounted to it.

Remaining issues:
- fix the `Username is invalid because files already exist for this user` problem
  - maybe use the nextcloud data root as the snoop collection root -- this means we lose all `uploads` data
  - maybe fork off `nextcloud` and remove that one `if` that checks for this "error"
  - maybe try and fix it in upstream https://github.com/nextcloud/server/issues/21144


gabriel-v:
Using a symlink doesn't work because we run into another error: symlinks are not supported when running `php occ files:scan`. I tried a workaround with `mount --bind` inside the docker container (which requires a privileged container with CAP_SYS_ADMIN). The mounts don't seem to propagate the way I want them to: inside the container `/data` and `/.../files` are the same, but outside of the container `$COLLECTIONS/uploads/data` remains empty (even if it's being mounted to /data`).

I think the simplest solution right now is to:
- fork nextcloud server code
- remove check for username
- remove check for not following symlinks when scanning data files

PR #283, opened 2020-09-02
Unbashify script inside nomad template

RazorBest:
A script inside a nomad template was using exclusive bash syntax, but it was run with /bin/sh. There were 2 solutions: run it with bash or change the script in accordance to POSIX. I chose the second method.

### DIFF
--- a/templates/_lib.hcl
+++ b/templates/_lib.hcl
@@ -124,7 +124,7 @@ ephemeral_disk {
 {%- macro set_pg_password_template(username) %}
   template {
     data = <<-EOF
-    #!/bin/sh
+    #!/bin/bash
     set -e
     sed -i 's/host all all all trust/host all all all md5/' $PGDATA/pg_hba.conf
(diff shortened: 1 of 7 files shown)


gabriel-v:
Sure, but the tests are failing.

I suggest we use `bash` everywhere instead of editing scripts to remove double [[: 

- change shebang line from sh to bash: https://github.com/liquidinvestigations/node/blob/cd9903d56bb3d295a9f16d0a7e3b3da80d0d8f57/templates/hoover.nomad#L49
- change program line from sh to bash: https://github.com/liquidinvestigations/node/blob/cd9903d56bb3d295a9f16d0a7e3b3da80d0d8f57/templates/hoover.nomad#L327
  - a search for `"sh"` and `/bin/sh` yields a lot of results; could you please remove `sh` from everywhere where possible?

gabriel-v (commented):
any progress on this?

gabriel-v:
merged to master

PR #284, opened 2020-09-16
Oauth2 proxy

RaduNichita:
Added integration of https://github.com/oauth2-proxy/oauth2-proxy to replace https://github.com/liquidinvestigations/authproxy/ as a proxy for request. It works on all services that we currently have, with one special condition for hypothesis

I have submitted some changes to the following repoos:

Hoover-Search -> https://github.com/liquidinvestigations/hoover-search/pull/179
Dokuwiki -> https://github.com/liquidinvestigations/liquid-dokuwiki/pull/4/
Oauth2 Liquid Provider -> https://github.com/liquidinvestigations/oauth2-proxy/pull/1/

The "count" variable in authproxy_group from templates/_lib.hcl has been removed, as the current proxy doesn't support having multiple upstream with the same path and hopefully fix the problems with the proxy from the past. 

From CRJI/EIC#464 && From CRJI/EIC#465

### DIFF
--- a/liquid_node/commands.py
+++ b/liquid_node/commands.py
@@ -25,22 +25,22 @@ CORE_AUTH_APPS = [
     {
         'name': 'authdemo',
         'vault_path': 'liquid/authdemo/auth.oauth2',
-        'callback': f'{config.app_url("authdemo")}/__auth/callback',
+        'callback': f'{config.app_url("authdemo")}/oauth2/callback',
     },
     {
         'name': 'hoover',
         'vault_path': 'liquid/hoover/auth.oauth2',
-        'callback': f'{config.app_url("hoover")}/__auth/callback',
+        'callback': f'{config.app_url("hoover")}/oauth2/callback',
     },
     {
         'name': 'dokuwiki',
         'vault_path': 'liquid/dokuwiki/auth.oauth2',
-        'callback': f'{config.app_url("dokuwiki")}/__auth/callback',
+        'callback': f'{config.app_url("dokuwiki")}/oauth2/callback',
     },
     {
         'name': 'rocketchat-authproxy',
         'vault_path': 'liquid/rocketchat/auth.oauth2',
-        'callback': f'{config.app_url("rocketchat")}/__auth/callback',
+        'callback': f'{config.app_url("rocketchat")}/oauth2/callback',
     },
     {
         'name': 'rocketchat-app',
@@ -50,7 +50,7 @@ CORE_AUTH_APPS = [
     {
         'name': 'nextcloud',
         'vault_path': 'liquid/nextcloud/auth.oauth2',
-        'callback': f'{config.app_url("nextcloud")}/__auth/callback',
+        'callback': f'{config.app_url("nextcloud")}/oauth2/callback',
     },
     {
         'name': 'codimd-app',
@@ -60,12 +60,12 @@ CORE_AUTH_APPS = [
     {
         'name': 'codimd-authproxy',
         'vault_path': 'liquid/codimd/auth.oauth2',
-        'callback': f'{config.app_url("codimd")}/__auth/callback',
+        'callback': f'{config.app_url("codimd")}/oauth2/callback',
     },
     {
         'name': 'hypothesis',
         'vault_path': 'liquid/hypothesis/auth.oauth2',
-        'callback': f'{config.app_url("hypothesis")}/__auth/callback',
+        'callback': f'{config.app_url("hypothesis")}/oauth2/callback',
     },
 ]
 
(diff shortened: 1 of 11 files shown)


liquid_node/commands.py
```diff
@@ -211,6 +211,56 @@ def check_system_config():
...
+    job_checks = {}
+    for service, checks in nomad.get_health_checks(spec):
+        if not checks:
+            log.warn(f'service {service} has no health checks')
+            continue
+        job_checks[service] = checks
+    return job_checks
+
+
+def populate_secrets(vault_secret_keys, core_auth_apps, liquid_job):
```
gabriel-v, line 227: Please add a docstring to this function, because the arguments here are pretty hard to descipher. Explain what steps this function goes through, what all the parameters do. Add examples too, if you would.

Here are some examples on the formatting you could use:
https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html#example-google

versions.ini
```diff
@@ -5,7 +5,7 @@ hoover-search = liquidinvestigations/hoover-search:0.5.9
 hoover-snoop2 = liquidinvestigations/hoover-snoop2:0.11.1
 hoover-ui = liquidinvestigations/hoover-ui:0.3.0
 hypothesis-h = liquidinvestigations/hypothesis-h:0.2.1
-liquid-authproxy = liquidinvestigations/authproxy:0.3.5
+liquid-authproxy = liquidinvestigations/oauth-proxy:liquid_provider
 liquid-core = liquidinvestigations/core:0.3.7
-liquid-dokuwiki = liquidinvestigations/liquid-dokuwiki:0.0.4
+liquid-dokuwiki = liquidinvestigations/liquid-dokuwiki:oauth2_path
 liquid-nextcloud = liquidinvestigations/liquid-nextcloud:0.2.3
```
gabriel-v, line 11: i remember you making pull requests in other apps, please add those here too

templates/_lib.hcl
```diff
@@ -71,18 +71,34 @@ ephemeral_disk {
...
+            OAUTH2_PROXY_CLIENT_SECRET = {{.Data.client_secret | toJSON }}
+          {{- end }}
+          {{- with secret "liquid/${name}/cookie" }}
+            OAUTH2_PROXY_COOKIE_SECRET = {{.Data.cookie | toJSON }}
+          {{- end }}
+            OAUTH2_PROXY_EMAIL_DOMAINS = *
+            OAUTH2_PROXY_HTTP_ADDRESS = "0.0.0.0:5000"
+            OAUTH2_PROXY_PROVIDER = "liquid"
+            OAUTH2_PROXY_REDIRECT_URL = "${config.liquid_http_protocol}://${name}.${config.liquid_domain}/oauth2/callback"
+            OAUTH2_PROXY_REDEEM_URL = "${config.liquid_http_protocol}://{{key "liquid_domain"}}/o/token/"
```
gabriel-v, line 85: please bring the correct config here and re-test it on https

templates/hypothesis.nomad
```diff
@@ -387,6 +387,7 @@ job "hypothesis" {
       host='hypothesis.' + liquid_domain,
       upstream='hypothesis',
       user_header_template="acct:{}@" + liquid_domain,
```
gabriel-v, line 389: remove/rename the macro arguments here too

gabriel-v (changes requested):

liquid_node/commands.py
```diff
@@ -211,6 +211,66 @@ def check_system_config():
...
+            'password': config.ci_docker_password,
+        })
+
+    ensure_secret('liquid/rocketchat/adminuser', lambda: {
+        'username': 'rocketchatadmin',
+        'pass': random_secret(64),
+    })
+
+    for name in config.ALL_APPS:
+        ensure_secret(f'liquid/{name}/cookie', lambda: {
```
gabriel-v, line 257: Why do we need a separate key structure for these? Let's use the hardcoded list in `vault_secret_keys` (maybe make it a global?) -- we need to have these in the same place, more `for app in apps: write out this new chain of secrets`  aren't helping the readability of this code...

This also means you'll have to use `.secret_key` instead of `.secrets` in nomad templates.

liquid_node/commands.py
```diff
@@ -381,17 +399,17 @@ def alloc(job, group):
...
 @click.argument('name')
-@click.argument('args', nargs=-1)
+@click.argument('args', nargs=-1, type=str)
 def shell(name, args):
     """Open a shell in a docker container addressed as JOB:TASK"""
     docker.shell(name, *args)
 
 
-@liquid_commands.command()
+@liquid_commands.command(context_settings={"ignore_unknown_options": True})
```
gabriel-v, line 410: I think I added these to master, rebase?

gabriel-v (changes requested):
only some small changes to go

templates/_lib.hcl
```diff
@@ -105,12 +121,12 @@ ephemeral_disk {
...
         ]
         check {
-          name = "http"
+          name = "ping"
           initial_status = "critical"
           type = "http"
-          path = "/__auth/logout"
-          interval = "6s"
-          timeout = "3s"
+          path = "/ping"
```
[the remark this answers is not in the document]
RaduNichita: yeah, no problems

liquid_node/commands.py
RaduNichita: If we put those key in the vault_secret_keys, a 256 - bits secrets would be generated, which is not supported by the proxy (64bits is the current maximum at 29/09/2020)

gabriel-v:
merged in commit https://github.com/liquidinvestigations/node/commit/46b005bb6786a87981f4d9bc0e4b51c8a96bf1b7

PR #290, opened 2020-10-26
Added reup

gabriel-v:
I have no idea what this service does, I remember it being some redirects to/from old IDs to new IDs.

Do we still need this branch here? Do we still need the functionality?


gabriel-v:
it's a local job now

PR #291, opened 2020-10-26
Browser Test Suite

gabriel-v:
WIP, thanks @RaduNichita


gabriel-v:
PR is badly out of date; we should make another one and only copy over these files:
- script to create a test user on liquid-core https://github.com/liquidinvestigations/node/blob/bf3e1d366d087013f075db944d55412dd8bc2d49/user_script.sh
- testing script https://github.com/liquidinvestigations/node/blob/bf3e1d366d087013f075db944d55412dd8bc2d49/test.py

PR #292, opened 2020-10-26
Run the UI with "next start" or "next dev" instead of serving static output of "next export"

gabriel-v:
Running node.js to serve the UI means it's possible to work on the code under auto-reload. Also, I didn't enjoy the weird path hacks happening.

This UI also:
- moves the search apI to /api/v0/
- removes rewriting of all responses in search

TODO:
- fix the busted UI... https://github.com/liquidinvestigations/hoover-ui/tree/run-server-in-node


gabriel-v:
merged

PR #308, opened 2021-03-19
Create a periodic backup script

salevajo:
Script and documentation added. First part of CRJI/EIC#689

### DIFF
--- a/docs/Maintenance.md
+++ b/docs/Maintenance.md
@@ -31,6 +31,24 @@ To restore a collection:
 ./liquid restore-collection /tmp/backup-today/collection-testdata new-collection-name
 ```
 
+### Create periodic backups
+
+The [./bin/periodic-backup.sh](../bin/periodic-backup.sh) script has one positional `--dir` and optional arguments:
+```
+Usage: ./bin/periodic-backup.sh --dir exportdir [--rm] [--days N]
+  --dir      backup directory, a date based subfolder will be added, e.g. exportdir/YYYYMMDD-HHmm
+  --rm       remove old backups, default disabled
+  --days N   remove backups older than N days, default 60 days
+```
+
+Using `crontab -e` you can create periodic daily and weekly backups like this:
+```shell
+# Create a Liquid backup every day at 6am and keep old ones 7 days
+0 6 * * * /opt/node/bin/periodic-backup.sh --dir /storage/backup/daily --rm --days 7
+
+# Create a Liquid backup every week at 4am and remove old ones
+0 4 * * 1 /opt/node/bin/periodic-backup.sh --dir /storage/backup/weekly --rm
+```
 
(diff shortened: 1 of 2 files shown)


docs/Maintenance.md
```diff
@@ -44,6 +63,21 @@ Data will only be restored if archives that exist at that location. This means
...
+`uploads` collection having `sync = True` in [`liquid.ini`](../liquid.ini): 
+
+```ini
+# Collection corresponding to nextcloud uploads
+[collection:uploads]
+process = True
+sync = True
+```
+
+Uploads and app data will change often and therefore needs to be backed up on a regular base.
```
gabriel-v, line 79: "regular basis"

docs/Maintenance.md
```diff
@@ -44,6 +63,21 @@ Data will only be restored if archives that exist at that location. This means
 you can choose to only restore a certain application by placing its archives in
 an empty directory and running this command on it.
 
+### A possible backup strategy
+
+Assuming most collections don't change after processing them for the first time,
+you can create a full backup once for each collection. Except the so called
+`uploads` collection having `sync = True` in [`liquid.ini`](../liquid.ini):
```
gabriel-v, line 70: link to `../examples/liquid.ini` here, since the `liquid.ini` doesn't exist.

docs/Maintenance.md
```diff
@@ -44,6 +63,21 @@ Data will only be restored if archives that exist at that location. This means
...
 an empty directory and running this command on it.
 
+### A possible backup strategy
+
+Assuming most collections don't change after processing them for the first time,
+you can create a full backup once for each collection. Except the so called
+`uploads` collection having `sync = True` in [`liquid.ini`](../liquid.ini): 
+
+```ini
+# Collection corresponding to nextcloud uploads
```
gabriel-v, line 73: i wouldn't copy/paste the config for that here. Instead, i'd change the text to say "Except for the `uploads` collection or any other collection with `sync = True` in liquid.ini, because the data for these is expected to change often."

gabriel-v (changes requested):
Script looks good.

Please fix a typo and rephrase the docs a bit to remove some duplication.

docs/Maintenance.md
gabriel-v: to replace the copy/pasted code section, i'd make a permalink like this: https://github.com/liquidinvestigations/node/blob/40963726bf79d3318496572e41f93543c93132f3/examples/liquid.ini#L252-L256

that way, you highlight the correct lines, while also having the selection work when we change that file.

salevajo: Changed and directly linked into https://github.com/liquidinvestigations/node/blob/40963726bf79d3318496572e41f93543c93132f3/examples/liquid.ini#L252-L256

PR #311, opened 2021-04-16
Thumbnail service

k-jell:
Closes CRJI/EIC#711


production-versions.ini
```diff
@@ -11,3 +11,4 @@ liquid-dokuwiki = liquidinvestigations/liquid-dokuwiki:0.1.0
 liquid-nextcloud = liquidinvestigations/liquid-nextcloud:0.2.3
 rocketchat = rocketchat/rocket.chat:3.9.1
 rocketchat-mongo = mongo:4.4
+thumbnail-service = fpurchess/preview-service:latest
```
gabriel-v, line 14: Let's link some specific version instead of `latest`. Using the `latest` tag means any breaking change they ever add will break our apps too. That's what tags are for.

If they don't tag their containers, you can use a `sha3`  docker image link instead.

staging-versions.ini
```diff
@@ -11,3 +11,4 @@ liquid-dokuwiki = liquidinvestigations/liquid-dokuwiki:release
 liquid-nextcloud = liquidinvestigations/liquid-nextcloud:0.2.3
 rocketchat = rocketchat/rocket.chat:3.9.1
 rocketchat-mongo = mongo:4.4
+thumbnail-service = fpurchess/preview-service:latest
```
gabriel-v, line 14: same as above: don't use `latest`

testing-versions.ini
```diff
@@ -11,3 +11,4 @@ liquid-dokuwiki = liquidinvestigations/liquid-dokuwiki:master
 liquid-nextcloud = liquidinvestigations/liquid-nextcloud:0.2.3
 rocketchat = rocketchat/rocket.chat:3.9.1
 rocketchat-mongo = mongo:4.4
+thumbnail-service = fpurchess/preview-service:latest
```
gabriel-v, line 14: same as above: don't use `latest`

liquid_node/configuration.py
```diff
@@ -160,6 +160,8 @@ def __init__(self):
         self.tika_count = self.ini.getint('liquid', 'tika_count', fallback=1)
         self.tika_memory_limit = self.ini.getint('liquid', 'tika_memory_limit', fallback=800)
 
+        self.thumbnail_memory_limit = self.ini.getint('liquid', 'thumbnail_memory_limit', fallback=800)
```
gabriel-v, line 163: is it really filling up 800MB of ram?

Take a look at the container RAM usage like this:
- visit http://10.66.60.1:9990/grafana
- visit the "docker stats" dashboard and look for the memory graph
- click your container name on the list
- observe max RAM usage while you re-process testdata with it

liquid_node/configuration.py
```diff
@@ -160,6 +160,8 @@ def __init__(self):
...
+
```
gabriel-v, line 164: Please add another parameter for the `count` (like we have with Tika, default 1). This is needed to scale on big machines with tens of CPUs, or clusters with many small machines.

templates/hoover-deps.nomad
```diff
@@ -351,6 +351,68 @@ job "hoover-deps" {
...
+          name = "http"
+          initial_status = "critical"
+          type = "http"
+          path = "/"
+          interval = "${check_interval}"
+          timeout = "${check_timeout}"
+        }
+        check_restart {
+          limit = 5
+          grace = "480s"
```
gabriel-v, line 408: Did you ever see it fail like this? This is needed when the container somehow enters a failure state without the container dying. For example, Tika and Rocketchat both have this behavior where they error out and start rejecting all HTTP traffic, but the containers are still up.

Let's leave out the `check_restart` from here unless you've seen it fail without the container dying.

gabriel-v (changes requested):
Please add entry in `examples/liquid.ini` with config values and comments.

gabriel-v (approved):

gabriel-v:
@k-jell please rebase and resolve conflicts

liquid_node/configuration.py
```diff
@@ -160,6 +160,9 @@ def __init__(self):
         self.tika_count = self.ini.getint('liquid', 'tika_count', fallback=1)
         self.tika_memory_limit = self.ini.getint('liquid', 'tika_memory_limit', fallback=800)
 
+        self.thumbnail_count = self.ini.getint('liquid', 'thumbnail_count', fallback=1)
```
gabriel-v, line 163: i'd like these in the `[snoop]` category alongside the
```
        self.snoop_pdf_preview_enabled = self.ini.getboolean('snoop', 'pdf_preview_enabled', fallback=False)
```

setting I added below.

Yes, the `tika_count` and friends should also be moved into that category, but that's a breaking change that's not worth it at the moment.

testing-versions.ini
```diff
@@ -11,4 +11,5 @@ liquid-dokuwiki = liquidinvestigations/liquid-dokuwiki:master
 liquid-nextcloud = liquidinvestigations/liquid-nextcloud:0.2.3
 rocketchat = rocketchat/rocket.chat:3.9.1
 rocketchat-mongo = mongo:4.4
-pdf-preview = thecodingmachine/gotenberg:6
+thumbnail-service = fpurchess/preview-service@sha256:1c61b92953d809dcb8b8133dc0353ae59449e640197fd4a139c2faa00fe74499
+pdf-preview = thecodingmachine/gotenberg:6
```
gabriel-v, line 15: please configure your editor to save newline at end of file

templates/hoover-workers.nomad
```diff
@@ -76,6 +77,7 @@ job "hoover-workers" {
       env {
         SNOOP_ES_URL = "http://{% raw %}${attr.unique.network.ip-address}{% endraw %}:9990/_es"
         SNOOP_TIKA_URL = "http://{% raw %}${attr.unique.network.ip-address}{% endraw %}:9990/_tika/"
+        SNOOP_THUMBNAIL_URL = "http://{% raw %}${attr.unique.network.ip-address}{% endraw %}:9990/_thumbnail/"
```
gabriel-v, line 80: Let's make this one optional too, with some `_enabled` config flag like the one for PDF previews below. By default it's false to make it backwards compatible.

templates/hoover-workers.nomad
```diff
@@ -61,6 +61,7 @@ job "hoover-workers" {
           # exec tail -f /dev/null
           if  [ -z "$SNOOP_TIKA_URL" ] \
                   || [ -z "$SNOOP_DB" ] \
+                  || [ -z "$SNOOP_THUMBNAIL_URL" ] \
```
gabriel-v, line 64: You don't need to add this to this `if` because it's not prone to disappear. This env is a constant (node IP + some path), as opposed to some others that originate from some `range service` loop that may sometimes have 0 iterations through it (e.g. when the database restarts). This means the env is always going to be set.

gabriel-v (changes requested):
Some small changes to make around the settings

templates/hoover-deps.nomad
```diff
@@ -405,6 +405,67 @@ job "hoover-deps" {
   }
   {% endif %}
 
+  group "thumbnail-service" {
```
gabriel-v, line 408: Please surround this with an `if` because setting `count = 0` raises an error in Nomad, and so we can't avoid running this service that we want to make optional.

examples/liquid.ini
```diff
@@ -83,6 +83,13 @@ two_factor_auth = false
 # Configure replication factor for tika.
 # The default is 1.
 ;tika_count = 3
+;
+# Configure memory limits for Thumbnail Service
+# The value is in MB. The default is:
+;thumbnail_memory_limit = 900
```
gabriel-v, line 89: document the `enabled` value first

templates/hoover-deps.nomad
```diff
@@ -405,6 +405,67 @@ job "hoover-deps" {
...
+    count = ${config.thumbnail_count}
+
+    ${ continuous_reschedule() }
+    ${ group_disk() }
+
+    task "thumbnail" {
```
gabriel-v, line 414: let's try to keep all the names predictable: the group is `thumbnail-service`, the task name is `thumbnail`, the fabio path is `_thumbnail`, but the key to get the image is `thumbnail-service` once again. Pick one descriptive name and use it all over, for example `thumbnail-generator`

liquid_node/configuration.py
```diff
@@ -203,6 +203,9 @@ def __init__(self):
         self.snoop_pdf_preview_count = self.ini.getint('snoop', 'pdf_preview_count', fallback=1)
         self.snoop_pdf_preview_memory_limit = self.ini.getint('snoop', 'pdf_preview_memory_limit',
                                                               fallback=900)
+        self.snoop_thumbnails_enabled = self.ini.getboolean('snoop', 'thumbnails_enabled', fallback=False)
```
gabriel-v, line 206: let's have these 3 variables have the same prefix `snoop_thumbnails_*` or `snoop_thumbnail_generator_*`

I'd like the prefix to be the same as the service group/task/image/fabio routes, see comment below

gabriel-v (changes requested):
Some superficial refactoring to do, unifying variable names, configs and such

gabriel-v (approved):

PR #316, opened 2021-07-14
Update elasticsearch to 7.13.2

k-jell:
closes CRJI/EIC#731


k-jell:
https://github.com/liquidinvestigations/node/blob/aef1d40999ca9f0e0e25499484249e2a3f38d44f/templates/hoover-deps.nomad#L41

We probably need to set the `discovery.seed_hosts` instead of using this, but I didn't know which value. (https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-discovery-hosts-providers.html)

templates/hoover-deps.nomad
```diff
@@ -55,7 +56,7 @@ job "hoover-deps" {
 
   group "es-master" {
     ${ continuous_reschedule() }
-    ${ group_disk() }
+    ${ group_disk(200) }
```
gabriel-v, line 59: why'd you need this?

gabriel-v (commented):
Yeah, don't set `single-node` because we want to be running in multi-node.

Try to give `discovery.seed_hosts` the IP of the master node only

If that doesn't work out, we can list there the IPs of all the different nodes, but that might not work, because nomad will cascade restart all other services when one of them is restarted - not what we want at all

Also remember to increase the timeouts to something like 10 retries of 45s each, so we end up at around 7min in total

templates/hoover-deps.nomad
k-jell: I had an error related to disk space which I thought was resolved by this. Something like log size exceeded or whatever. But apparently it wasn't related and resolved by something else. Removed it now.

templates/hoover-deps.nomad
```diff
@@ -38,6 +38,7 @@
         xpack.monitoring.collection.index.stats.timeout = "30s"
         xpack.monitoring.collection.index.recovery.timeout = "30s"
         xpack.monitoring.history.duration = "32d"
+        discovery.seed_hosts = "{% raw %}${attr.unique.network.ip-address}{% endraw %}"
```
gabriel-v, line 41: Please don't set this here, since it's also overridden in the job descriptions for `es-data`. If you need this config on the master node, set it explicitly there, not in this macro.

gabriel-v (changes requested):

PR #318, opened 2021-10-14
Create user in all apps after it has been created in core

k-jell:
closes CRJI/EIC#745

### DIFF
--- a/liquid_node/jobs/__init__.py
+++ b/liquid_node/jobs/__init__.py
@@ -21,10 +21,6 @@ def set_volumes_paths(substitutions={}):
 
     from ..configuration import config
 
-    def get_str(str):
-        return str
-
-    substitutions['getstr'] = get_str
     substitutions['config'] = config
     substitutions['liquid_domain'] = config.liquid_domain
     substitutions['liquid_volumes'] = config.liquid_volumes
--- a/liquid_node/jobs/__init__.py
+++ b/liquid_node/jobs/__init__.py
@@ -21,6 +21,10 @@ def set_volumes_paths(substitutions={}):
 
     from ..configuration import config
 
+    def get_str(str):
+        return str
+
+    substitutions['getstr'] = get_str
     substitutions['config'] = config
     substitutions['liquid_domain'] = config.liquid_domain
     substitutions['liquid_volumes'] = config.liquid_volumes
--- a/templates/hypothesis.nomad
+++ b/templates/hypothesis.nomad
@@ -147,6 +147,35 @@ job "hypothesis" {
           destination = "local/usersync.py"
       }
 
+      template {
+        data = <<-EOF
+          #!/usr/bin/env python3
+          import sys, subprocess, secrets
+          class SafeString(str):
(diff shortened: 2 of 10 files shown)


templates/hypothesis.nomad
```diff
@@ -147,6 +147,37 @@ job "hypothesis" {
           destination = "local/usersync.py"
       }
 
+      template {
+        data = <<-EOF
+          #!/usr/bin/env python3
```
gabriel-v, line 152: Please move this into a separate file and include it here.

Take care with whitespace; check [here](https://github.com/liquidinvestigations/node/blob/60cee552d59745bc6b93d53c67253c08b38db35f/templates/rocketchat-migrate.nomad#L28-L29) for an example of how to include.

Throw all the scripts in a `scripts` subfolder, we'll refactor all the others in the future to separate scripts from nomad files and keep them in that subfolder.

templates/liquid-createuser.nomad
```diff
@@ -0,0 +1,85 @@
...
+      config {
+        command = "sh"
+        args    = ["local/createuser.sh"]
+      }
+
+      template {
+        destination = "local/create_rocket_user.py"
+        perms = "755"
+        data = <<-EOF
+        import requests
```
gabriel-v, line 25: As above, please move this script into a separate file and include it here.

templates/liquid-createuser.nomad
```diff
@@ -0,0 +1,85 @@
...
+        requests.post(url + '/api/v1/logout', headers=headers)
+        EOF
+      }
+
+      template {
+        destination = "local/createuser.sh"
+        perms = "755"
+        data = <<-EOF
+          #!/bin/sh
+          apt-get install pip
```
gabriel-v, line 59: As above, please move this script into a separate file and include it here.

templates/liquid-createuser.nomad
```diff
@@ -0,0 +1,85 @@
...
+          pip install requests
+          set -ex
+          liquid_domain="${liquid_domain}"
+          echo "Creating user ${getstr('${NOMAD_META_USERNAME}')} for all enabled apps."
```
gabriel-v, line 63: Please use `raw` tags like [elsewhere](https://github.com/liquidinvestigations/node/blob/061a57f9dc2617e40f7ac5254ad609bd2bc1a676/templates/hoover-workers.nomad#L16) instead of adding a shortcut to `str` here.

liquid_node/jobs/__init__.py
```diff
@@ -21,6 +21,10 @@ def set_volumes_paths(substitutions={}):
 
     from ..configuration import config
 
+    def get_str(str):
```
gabriel-v, line 24: Please remove all this in favor of `raw` and `endraw` tags, see comment below.

Also, a wrapper would not be needed here, you could have done `substititutions['getstr'] = str`, or even better, keep the id as `str` instead of `getstr`.

gabriel-v (changes requested):

PR #323, opened 2022-02-11
Ner integration

k-jell:
closes CRJI/EIC#297


examples/liquid.ini
```diff
@@ -251,6 +251,21 @@ wait_green_count = 6
 # - `image_classification_classify_images_model`   - Set the model used for image classification. Default is "mobilenet".
 #                                                    Options are "mobilenet", "resnet", "inception" and "densenet".
 # for documentation of the image classification settings see: https://github.com/liquidinvestigations/image-classification-service
+# NLP service for named entity recocnition
```
gabriel-v, line 254: typo `recocnition`

examples/liquid.ini
```diff
@@ -251,6 +251,21 @@ wait_green_count = 6
...
+# - `nlp_entity_extraction_enabled` - Setting to enable entity extraction. Default is "false".
+# - `nlp_language_detection_enabled` - Setting to enable language detection. Default is "false".
+# - `nlp_count` - Set the number of containers to run. Default is 2
+# - `nlp_fallback_language` - language that language detection returns if no
+#                 language can be determined by service
```
gabriel-v, line 259: The style of the lines above this is:
- sentence starts with uppercase, ends with .
- overflow lines align with previous `-`

Let's try to keep that style here too

examples/liquid.ini
```diff
@@ -277,6 +308,26 @@ image_classification_object_detection_enabled = false
 
 
 
+;[nlp_service]
```
gabriel-v, line 311: Let's not add a new section for this. I see it's not read in the configuration file, so this is left over dead code i suppose.

templates/hoover-deps.nomad
```diff
@@ -524,7 +524,70 @@ job "hoover-deps" {
   }
   {% endif %}
 
-  group "search-rabbitmq" {
```
gabriel-v, line 527: Rebase error: keep the group name unchanged `search-rabbitmq`...

templates/hoover-deps.nomad
```diff
@@ -524,7 +524,70 @@ job "hoover-deps" {
...
+          path = "/config"
+          interval = "${check_interval}"
+          timeout = "600s"
+        }
+      }
+    }
+  }
+  {% endif %}
+
+  group "rabbitmq" {
```
gabriel-v, line 590: ... renamed into this: from `search-rabbitmq`

templates/hoover-deps.nomad
```diff
@@ -524,7 +524,70 @@ job "hoover-deps" {
...
+        name = "hoover-nlp-service"
+        tags = ["fabio-/_nlp strip=/_nlp"]
+        port = "nlp"
+        check {
+          name = "http"
+          initial_status = "critical"
+          type = "http"
+          path = "/config"
+          interval = "${check_interval}"
+          timeout = "600s"
```
gabriel-v, line 583: Why do we need such a huge timeout? Is the `/config` route slow? 

Let's put the usual value `${check_timeout}` here, and if it fails, use a faster route for `/config`, or set up more workers/threads for gunicorn

templates/hoover-deps.nomad
```diff
@@ -524,7 +524,70 @@ job "hoover-deps" {
...
+        NLP_SERVICE_FALLBACK_LANGUAGE = "${config.snoop_nlp_fallback_language}"
+        NLP_SPACY_TEXT_LIMIT = "${config.snoop_nlp_spacy_text_limit}"
+        GUNICORN_WORKERS = ${config.snoop_nlp_gunicorn_workers}
+        GUNICORN_THREADS = ${config.snoop_nlp_gunicorn_threads}
+        EXTRACT_ENTITIES = "${config.snoop_nlp_entity_extraction_enabled}"
+        DETECT_LANGUAGE = "${config.snoop_nlp_language_detection_enabled}"
+      }
+      resources {
+        memory = ${config.snoop_nlp_memory_limit}
+        cpu = 1500
```
gabriel-v, line 567: keep this lower, something like 300-500. This doesn't enforce a limit, it just stops us from starting more containers than the detected cpu powoer, and we don't want to be limited like that.

gabriel-v (changes requested):
How does snoop know if `snoop_nlp_language_detection_enabled` has been set? You're only sending the URL to snoop containers.

Same question for `spacy_text_limit` -- how does snoop know when to truncate the text, or to split it into separate requests? This is info the caller I think should have.

Are you hitting your `/config` endpoint and finding out from there?

If so, I think it should be best to move all those specific settings into more environment variables on the snoop containers, instead of doing API calls every request / on startup

templates/hoover-deps.nomad
k-jell: It was this huge, because the nlp-service needs so long to set up and download the models. The `/config` route is only available afterwards.

PR #333, opened 2022-07-01
improve collection name validation

k-jell:
closes CRJI/EIC#860

### DIFF
--- a/liquid_node/configuration.py
+++ b/liquid_node/configuration.py
@@ -486,12 +486,12 @@ class Configuration:
             self.ini.getboolean('apps', app_name, fallback=strtobool(self.default_app_status))
 
     @classmethod
-    def _validate_collection_name(self, name):
+    def _validate_collection_name(cls, name):
 
-        # collection names are restricted by minio and elasticsearch
+        # collection names are restricted by minio (s3) and elasticsearch
         # for reference see:
         # https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-create-index.html#indices-create-api-path-params
-        # https://docs.abp.io/en/abp/latest/Blob-Storing-Minio#options
+        # https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html
 
         search_forbidden_char = re.compile(r'[^a-z0-9-]').search
 
--- a/liquid_node/configuration.py
+++ b/liquid_node/configuration.py
@@ -1,4 +1,5 @@
 import sys
+import re
 import logging
 import configparser
 import os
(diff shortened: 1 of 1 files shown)


liquid_node/configuration.py
```diff
@@ -486,16 +487,25 @@ def is_app_enabled(self, app_name):
...
     @classmethod
     def _validate_collection_name(self, name):
-        if not name.islower():
+
+        # collection names are restricted by minio and elasticsearch
+        # for reference see:
+        # https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-create-index.html#indices-create-api-path-params
+        # https://docs.abp.io/en/abp/latest/Blob-Storing-Minio#options
+
+        search_forbidden_char = re.compile(r'[^a-z0-9-]').search
```
gabriel-v, line 496: Wait -- uppercase is OK, right? underscores? dots?

In prod, we have loads of uppercase collection names with dashes. We don't have any dots or underscores.

liquid_node/configuration.py
```diff
@@ -486,16 +487,25 @@ def is_app_enabled(self, app_name):
 
     @classmethod
     def _validate_collection_name(self, name):
```
gabriel-v, line 489: First argument for `@classmethod` should be `cls` not `self`, to signify we have a class type, not an object. Please fix this in this patch.

gabriel-v (changes requested):

liquid_node/configuration.py
k-jell: I didn't get the correct docs there. It should be the S3 bucket rules that apply? (https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html).
Still underscores and capital letters seem to be not allowed. Dots are allowed but then we would have to handle some special cases so maybe we refrain from allowing them at all.

gabriel-v: Here's our confusion described here:
https://github.com/minio/minio/issues/9088#issuecomment-595227401

> Existing directories on your disk in standalone mode are already supported to have _ or capitalized names.
> We do not support creating new buckets in this manner because of AWS S3 DNS requirements.

They accept the capitalized names when they already exist on disk. We have some of those.

If you add the restriction here, the service will refuse to deploy the already existing collections.

We don't want any special logic here relatetd to old vs. new collections.

We also don't want to rename all of the collections into lowercase, because that will break the existing links for the old collections, and we also don't have any kind of collection aliases or redirects.

Things to do:
- add uppercase to the check
-

gabriel-v (changes requested):

gabriel-v (approved):
lgtm

PR #335, opened 2022-07-20
add bash script to import tags

k-jell:
closes CRJI/EIC#876


import-tags.sh
```diff
@@ -0,0 +1,15 @@
...
+do
+    case "${flag}" in
+        c) collection="${OPTARG}" ;;
+        t) file="${OPTARG}" ;;
+        p) public="-p" ;;
+        u) user="${OPTARG}"
+    esac
+
+uuid=$(./liquid dockerexec hoover:search ./manage.py getuuid $user | tail -1)
+./liquid dockerexec hoover:snoop -c $collection -t $file --uuid $uuid --user $user $public
```
gabriel-v, line 15: how do you feed the $file inside the hoover:snoop container? we have no shared mounts and we don't want to add any.

I think you need to change all the scripts to accept the CSV as std in.

import-tags.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash -e
+
+public=''
+
+while getopts c:t:pu: flag
```
gabriel-v, line 5: getopts are complicated, shouldn't this be `c:t:p:u`? Could we also add long options here that correspond to the snoop ones?

import-tags.sh
```diff
@@ -0,0 +1,15 @@
...
+do
+    case "${flag}" in
+        c) collection="${OPTARG}" ;;
+        t) file="${OPTARG}" ;;
+        p) public="-p" ;;
+        u) user="${OPTARG}"
+    esac
+
+uuid=$(./liquid dockerexec hoover:search ./manage.py getuuid $user | tail -1)
```
gabriel-v, line 14: without `set -o pipefail` the `|tail -1` will make any possible errors in `getuuid` command fail silently, and we'll get an empty string here and no error.

please add `set -o pipefail` at the top of the script, at line 2. 

Also, run some manual tests for what happens when bad username is given; script should halt at this line.

gabriel-v (changes requested):
small questions related to the script

import-tags.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash -e
```
gabriel-v, line 1: if you added the python script, do  you still need this script?

gabriel-v (changes requested):

gabriel-v:
code looks good, after cleaning up the unused script, please squash & merge into master

PR #338, opened 2022-08-02
Delete users

k-jell:
closes CRJI/EIC#804


templates/scripts/deleteuser.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
+apt-get install pip
+pip install requests
+set -ex
+liquid_domain="${liquid_domain}"
```
gabriel-v, line 5: same as above: don't use any special Nomad or Jinja syntax.

In this bash script, I think it's best to simply use `$1` and `$2`.

The `${XXX}` syntax in Bash is problematic because it will be picked up by Jinja templating engine instead of Bash. `$1` and `$2`, which you can set to some named variables `$SOME_THING_1` and `$SOME_THING_2` do not have that problem.

templates/scripts/delete_rocket_user.py
```diff
@@ -0,0 +1,27 @@
...
+data = '{"username": "%s"}' % (username)
+# call the rocketchat api to delete user
+response = requests.post(url + 'api/v1/users.delete', headers=headers, data=data)
+if response.status_code != 200:
+    print('An error occured!')
+    print('Got the following response from rocketchat:')
+    print(response.content)
+else:
+    print(f'Deleted Rocketchat user: {username}.')
+requests.post(url + '/api/v1/logout', headers=headers)
```
gabriel-v, line 27: wait, so you log out after you delete the user? That seems suspect, don't you want it the other way around?

gabriel-v (changes requested):
Don't put jinja / consul template tags in the scripts. If you do that, they cannot be reused outside of your specific templating scheme, or debugged easily.

Second, make sure the templating engines won't work in the scripts. This causes confusion when you want to use `${variable}` in Bash, it is interpreted by Jinja. Similar problems with Consul-Template, which has the exact same template tags.

---

Here's how you disable templating tags in Consul-Template in the Nomad config: override the `{{` and `}}` delimiter tags

https://github.com/liquidinvestigations/node/blob/9535083a5aed39caf673172da2b3936ab0cf3b6f/templates/monitoring.nomad#L184-L185

---

And here's how to disable the Jinja tags interpreting, if you really need to:
https://stackoverflow.com/a/9769454

You should add that new function in the globals here:
https://github.com/liquidinvestigations/node/blob/16e0157c4af160ce7c22ccf23a40405f58fed881/liquid_node/jobs/__init__.py#L118

templates/scripts/delete_rocket_user.py
k-jell: yes because it logs out the admin user which is used to delete another account

gabriel-v (approved):
lgtm

PR #339, opened 2022-08-05
add dummy task if workers disabled

k-jell:
closes CRJI/EIC#894


templates/hoover-workers.nomad
```diff
@@ -231,5 +232,19 @@ job "hoover-workers" {
...
+  group "dummy" {
+    task "dummy" {
+      ${ task_logs() }
+
+      driver = "docker"
+      config {
+        image = "${config.image('pause')}"
+        labels {
+          liquid_task = "dummy"
+             }
```
gabriel-v, line 245: what's up with the whitespace here?

these don't look aligned to me

PR #341, opened 2022-10-28
Resumable uploads

k-jell:
closes CRJI/EIC#601

### DIFF
--- a/liquid_node/configuration.py
+++ b/liquid_node/configuration.py
@@ -171,6 +171,9 @@ class Configuration:
 
         self.hoover_ratelimit_user = self.ini.get('liquid', 'hoover_ratelimit_user', fallback='100,13')
 
+        self.hoover_tus_upload_dir = self.ini.get('liquid', 'hoover_tus_upload_dir', fallback='/tmp/tus/uploads')
+        self.hoover_tus_files_dir = self.ini.get('liquid', 'hoover_tus_files_dir', fallback='/tmp/tus/files')
+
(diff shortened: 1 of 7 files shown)


liquid_node/configuration.py
```diff
@@ -171,6 +171,11 @@ def __init__(self):
 
         self.hoover_ratelimit_user = self.ini.get('liquid', 'hoover_ratelimit_user', fallback='100,13')
 
+        self.hoover_tus_upload_dir = self.ini.get('liquid', 'hoover_tus_upload_dir',
```
gabriel-v, line 174: 1. this doesn't need be customizable
2. if this means a path on host, /tmp is a bad idea, since that is usually small and/or hosted on ramdisk

Let's have the upload temp path be local to the container, in the `/tmp` that is already mounted by Nomad to a disk folder using `tmpfs`.

This brings about another problem: do we need this temp dir to be shared between different servers if the hoover HTTP container count is > 1 ?

We don't want to use s3fs to mount between different containers, because that system is not consistent, and you get "file not found" just after uploading a file...

So:
- remove the path configs
- use default path in /tmp inside container
- comment out the `count` configs for snoop http web server container here: https://github.com/liquidinvestigations/node/blob/047dad5c105f99fe6e584d527584787d4c5da89f/templates/hoover.nomad#L236
- run a manual test where you upload a 10GB file and see how big the temp file becomes while uploading (using `du` in a `for` loop): is it going to peak at 10GB? 20GB?

gabriel-v:
flake8 was broken on master, please rebase

PR #355, opened 2023-06-23
[del] doc pages

thifranc:
rocketchat+hoover as merged in wiki-doc
test+nextcloud as irrelevant-insufficient

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unreleased
 
 ### Upgrade Notes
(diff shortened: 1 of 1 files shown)


thifranc:
I've also removed Cluster.md file as info contained in it is, to me, either obsolete or available at https://github.com/liquidinvestigations/cluster

gabriel-v:
Sure, we got 2 problems:
- readme local link to docs/Cluster is going to be 404
- the info kept there is not shown anywhere else; if you look closely, it's about interfacing liquid.ini vs. cluster.ini, not just about setting up the cluster. Do we have this cloned into a wiki page too? If so, pls paste its link here

PR #357, opened 2023-09-14
add env to disable dynamic collection admin during startup

k-jell:
required for this: https://github.com/liquidinvestigations/hoover-snoop2/pull/488


gabriel-v:
if you put the export [in the runserver script](https://github.com/liquidinvestigations/hoover-snoop2/blob/master/runserver), we can keep this complication in that repo, and away from this one - let's put it there

PR #358, opened 2023-10-09
[fix] cleaner output message for backup handling

thifranc:
minor fix to have a slightly more clear message in case restoring on already existing collection


gabriel-v:
fixed on commit in parallel

PR #359, opened 2024-01-12
[add] bbb greenlight in liquid

thifranc:
We have 3 more containers:

greenlight + postgres + redis, all of that for GREENLIGHT

Greenlight can be integrated within liquid-core OIDC as per noted in liquid.ini example.
We MUST use the liquid-core https://github.com/liquidinvestigations/core/tree/tmp-admin-show-oauth2 branch, to retrieve data credentials.

This greenlight is capable to start and make call to remote BBB servers, we hope to then provide BBB server within liquid, but for now, BBB server deployment+securing the connection is the task of sys-admins deploying liquid


thifranc:
By pushing the docs, I've just got aware that the actual PR is forcing BBB to be shipped, we should rather add BBB as an app ( as it's the case with like ';hoover = on' / ;codimd = on etc on the [apps] section of liquid.ini ).
That's actually not the case and it's a pity, @gabriel-v I'd be glad to add that change it you can point me towards the relevant piece of code.

gabriel-v:
https://github.com/liquidinvestigations/node/blob/bed0fa421592d8c4f6cb75fda20ca9f903ea2838/liquid_node/configuration.py#L55

https://github.com/liquidinvestigations/node/blob/bed0fa421592d8c4f6cb75fda20ca9f903ea2838/liquid_node/configuration.py#L64

https://github.com/liquidinvestigations/node/blob/bed0fa421592d8c4f6cb75fda20ca9f903ea2838/liquid_node/configuration.py#L74

docs/Bbb.md
```diff
@@ -0,0 +1,97 @@
...
+**This documentation will assume your BBB server has a unique public IP**, but you SHOULD protect your BBB server the same way your Liquid instance is, so if all traffic to Liquid is tunneled via a VPN, your traffic to BBB should do the same ; also, BBB software will try and detect itself its public IP, so the server is better with only one IP.
+If you move apart from that assumption (server with unique public IP) in any way, please refer to official BBB documentation, otherwise functional BBB server is not guaranteed.
+
+Install latest docker on your server :
+```bash
+apt-get update
+apt-get install ca-certificates curl
+install -m 0755 -d /etc/apt/keyrings
+curl -fsSL https://download.docker.com/linux/ubuntu/gpg -o /etc/apt/keyrings/docker.asc
+chmod a+r /etc/apt/keyrings/docker.asc
```
gabriel-v, line 32: https://get.docker.com | sudo bash

docs/Bbb.md
```diff
@@ -0,0 +1,97 @@
...
+as per the docker doc recommands, you can try and assure Docker is well installed by running :
+`docker run hello-world`
+
+Then to install latest BBB software (version 2.7), also as per the official doc recommands (read and change two variables before running) :
+Options used :
+`-e` is the email your will provide for being contacted regarding Letsencrypt SSL certificates, you can drop that option if generated your own certificates yourselves
+`-s` is the name your BBB server will be configured with
+`-v` is the BBB software version you're querying
+```
+wget -qO- https://raw.githubusercontent.com/bigbluebutton/bbb-install/v2.7.x-release/bbb-install.sh | bash -s -- -v focal-270 -s YOUR_SERVER_HOSTNAME -e email_for_letsencrypt_certbot@your_domain.org
```
gabriel-v, line 49: split in 2 commands

docs/Bbb.md
```diff
@@ -0,0 +1,97 @@
...
+```
+bbb_endpoint = YOUR_BBB_API_URL
+bbb_secret = YOUR_BBB_API_TOKEN
+```
+
+Then, you have to create a connect agent on the liquid admin interface, it's located at :
+`https://YOUR_LIQUID_FQDN/admin/oauth2_provider/application/`
+The info to fill in is as follows :
+```
+client type : 'confidential'
```
gabriel-v, line 84: let's automate the creation of SSO stuff

thifranc: when using 'core_oauth_apps' to automatically create the SSO app, I don't get to the right config, I lack having Algorithm: set as HMAC with SHA-2 256. ( and, skip authorization set as false ).

Could I (or anyone) add a commit into liquid-core ./liquidcore/home/management/commands/createoauth2app.py regarding this issue ? Just so that I can get something like :

```
        if options['noskip'] :
          app.skip_authorization = False
        else:
          app.skip_authorization = True

        if options['algo'] :
          app.algorithm = options['algo']
```
in https://github.com/liquidinvestigations/core/blob/c0b52b240f1305e1ed28acb74644d5a3d6d8b173/liquidcore/home/management/commands/createoauth2app.py#L22

Otherwise, I got into the following code, then error, within django-oauth-toolkit :
https://github.com/jazzband/django-oauth-toolkit/blob/40003dddda5b13f91b841ca267fd730487338c89/oauth2_provider/models.py#L258
aka "This application does not support signed tokens"

thifranc: Done and tested on my part, LGTM, code added on liquid-core https://github.com/liquidinvestigations/core/pull/211

docs/Bbb.md
```diff
@@ -0,0 +1,81 @@
...
+Now, to deploy liquid with the BBB frontend, you should get the credentials from your BBB server, to do so, connect to the BBB server and use the bbb-conf tool like this :
+`/usr/bin/bbb-conf --secret`
+that will give you the BBB URL and SECRET to fill in the liquid.ini file, copy-paste them as follows :
+```
+bbb_endpoint = YOUR_BBB_API_URL
+bbb_secret = YOUR_BBB_API_TOKEN
+```
+
+**IMPORTANT**
+You can now connect to the 'greenlight' (the BBB frontend) interface with your liquid users ! But the users HAVE to have a first name and surname with at least 2 characters. Otherwise, you will get an error.
```
gabriel-v, line 81: let's try and fix this by editing liquid-core to give BBB the username in place of the missing fields

examples/liquid.ini
```diff
@@ -149,6 +149,7 @@ two_factor_auth = false
 ;hoover_ui_async_search_error_multiplier = 2
 ;hoover_ui_async_search_error_summation = 60
 
+
```
gabriel-v, line 152: extra whitespace

liquid_node/configuration.py
```diff
@@ -72,6 +72,7 @@ class Configuration:
         'nextcloud': 'has a file share system and a contact list of users.',
         'rocketchat': 'is the old chat app; will remove shortly.',
         'matrix': 'is the new chat app',
+        'bbb': 'your visio app',
```
gabriel-v, line 75: visio i think is a francophone word, best call it "video conference" because "video chat" may have unprofessional connotations

liquid_node/configuration.py
```diff
@@ -633,6 +648,11 @@ def tag(name):
         if name in ['dokuwiki', 'nextcloud']:
             return tag('liquid-' + name)
 
+        if name == 'bbb':
+            bbb_gl = tag('bbb-gl')
+            bbb_redis = tag('bbb-redis')
+            return f'greenlight: {bbb_gl}, redis: {bbb_redis}'
```
gabriel-v, line 654: let's only display the greenlight version, these are user-facing version numbers, we don't want to have too many of them

templates/bbb.nomad
```diff
@@ -0,0 +1,217 @@
...
+        BIGBLUEBUTTON_SECRET = "${config.bbb_secret}"
+        {{- with secret "liquid/bbb/bbb.secret_key_base" }}
+        SECRET_KEY_BASE = {{.Data.secret_key | toJSON }}
+        {{- end }}
+        RELATIVE_URL_ROOT = "/"
+        {{- with secret "liquid/bbb/auth.oauth2" }}
+        OPENID_CONNECT_CLIENT_ID = {{.Data.client_id | toJSON }}
+        OPENID_CONNECT_CLIENT_SECRET = {{.Data.client_secret | toJSON }}
+        {{- end }}
+        OPENID_CONNECT_ISSUER="https://${liquid_domain}"
```
gabriel-v, line 169: what if this is hosted on HTTP (for example tor)?

IIRC webrtc doesn't even start connections without HTTPS, so it might not make sense to put the protocol vars here instead of the hard-coded `https`

templates/bbb.nomad
```diff
@@ -0,0 +1,217 @@
...
+          port "bbb-gl" {}
+        }
+      }
+
+      service {
+        name = "bbb-gl"
+        port = "bbb-gl"
+        tags = [
+        "traefik.enable=true",
+        "traefik.frontend.rule=Host:bbb.${liquid_domain}",
```
gabriel-v, line 204: Putting "greenlight" directly on the Traefik means people will see (and be able to attack/bruteforce) the login form of that app directly.

We usually try to wrap all possible HTTP/S traffic in the "authproxy" microservice, even for services with already have OAuth2 integration - this ensures browsers have valid login before even connecting them to the app.

Some services (like Matrix) aren't behind that proxy because they break external app functionality - is this something that we are interested here too? (for example BBB on tablets/phones?)

liquid_node/jobs/bbb.py
```diff
@@ -0,0 +1,23 @@
...
+        "liquid/bbb/bbb.postgres",
+        "liquid/bbb/bbb.secret_key_base",
+    ]
+    core_oauth_apps = [
+        {
+            'name': 'bbb',
+            'subdomain': 'bbb',
+            'vault_path': 'liquid/bbb/auth.oauth2',
+            'callback': '/auth/openid_connect/callback',
+            'algo': 'HS256',
```
gabriel-v, line 19: you added these `algo` and `noskip` flags here, but i don't see you actually using them with your[ `--algo` and `--noskip` ](https://github.com/liquidinvestigations/core/pull/214/files) arguments here -- https://github.com/liquidinvestigations/node/blob/bbb/liquid_node/commands.py#L212


This probably means this code won't run on a fresh installation, since it won't give the arguments to the initial oauth2 app creation. After fixing  all the suggestions in these PRs, please wipe all the data from your server (or spin up a new VM) and do the whole setup from scratch following your own docs - that should catch some more similar problems.

thifranc: indeed I missed pushing some code from liquid_node/commands.py, I'll add that right away.

templates/bbb.nomad
thifranc: nice catch, replaced with config.liquid_http_protocol, re-deployed, tested, ok now

liquid_node/configuration.py
thifranc: done

examples/liquid.ini
thifranc: I don't get it in my source code so don't know where that come from, either way I resolve, we can remove that later

docs/Bbb.md
thifranc: on the way to fix it, see the upcoming push for liquid-core, so I removed that part from docs, as will be obsolete when merged

templates/bbb.nomad
thifranc: I didn't quite get that part, I think I'll look at dokuwiki template-authproxy, and arrange a test from scratch with that.

thifranc: I tried to mimic config based on what's done with dokuwiki, but it doesn't work as is, maybe we'll need to tweak some files, but in the meantime, I think we can go with the actual version and make a following issue regarding that part

thifranc: resolved with https://github.com/liquidinvestigations/node/pull/361

liquid_node/configuration.py
```diff
@@ -250,6 +259,8 @@ def __init__(self):
         self.matrix_session_lifetime = self.ini.get('matrix', 'session_lifetime', fallback='168h')
         self.matrix_media_lifetime = self.ini.get('matrix', 'media_lifetime', fallback='365d')
         self.matrix_message_lifetime = self.ini.get('matrix', 'message_lifetime', fallback='365d')
+        self.bbb_endpoint = self.ini.get('bbb', 'bbb_endpoint', fallback='')
+        self.bbb_secret = self.ini.get('bbb', 'bbb_secret', fallback='')
```
gabriel-v, line 263: add an assert here (or lower) that checks both these params are set to non-empty values if BBB is on. Assert string should state that `[bbb]bbb_endpoint, _secret must be set when bbb is turned on`


Since this requires manual configuration, the default ini file will crash at boot. We don't want that to happen. Make bbb be off by default, even if `default_app_status = on` -- check `grist` and `prophecies` on how it's done (a small change in the `is_enabled` fn)

templates/bbb-migrate.nomad
```diff
@@ -0,0 +1,44 @@
...
+        args = ["/bin/sh", "/local/run.sh"]
+        volumes = [
+        ]
+      }
+
+        // DATABASE_URL = "postgres://greenlight:{{.Data.secret_key | toJSON }}@{{env "attr.unique.network.ip-address"}}:${config.port_bbb_pg}/greenlight_production"
+      template {
+        data = <<-EOF
+        {{- with secret "liquid/bbb/bbb.postgres" }}
+        PGPASSWORD={{.Data.secret_key | toJSON }} psql -p ${config.port_bbb_pg} -h {{env "attr.unique.network.ip-address"}} -U greenlight greenlight_production -c "update rooms_configurations set value=false from meeting_options where meeting_options.id=rooms_configurations.meeting_option_id and meeting_options.name='record';"
```
gabriel-v, line 35: this line is very long and we can shorten it:

- use \ and split the psql args
- use env vars with `export X=y`
- put both sql statements on same psql command separated by ;
- use multiline string buffer for the SQL  for example

```
USAGE=$(cat <<-END
    This is line one.
    This is line two.
    This is line three.
END
)

```

templates/bbb.nomad
```diff
@@ -0,0 +1,90 @@
...
+        volumes = [
+          "{% raw %}${meta.liquid_volumes}{% endraw %}/bbb/greenlight/data/:/usr/src/app/storage/",
+        ]
+        port_map {
+          bbb-gl = 3000
+        }
+        labels {
+          liquid_task = "bbb_greenlight"
+        }
+        memory_hard_limit = 2000
```
gabriel-v, line 64: user interfaces sometimes want more ram because of NPM / js things - let's give it here at least what it says on the box (8gb?)

templates/bbb-deps.nomad
```diff
@@ -0,0 +1,139 @@
...
+        attribute = "{% raw %}${meta.liquid_large_databases}{% endraw %}"
+        value     = "true"
+        weight    = 100
+      }
+
+      driver = "docker"
+      ${ shutdown_delay() }
+
+      config {
+        image = "postgres:15"
```
gabriel-v, line 31: since this image is duplicated in the "migrate" job, it's a good candidate for adding in the "default-versions.ini" file - call it something unique like "bbb-pg-15"

examples/liquid.ini
```diff
@@ -190,6 +190,7 @@ two_factor_auth = false
 ;dokuwiki = on
 ;codimd = on
 ;wikijs = on
+;bbb = off
```
gabriel-v, line 193: see comment below about making BBB work like grist & prophecies (default off, because requires manual config) -- also move this line to the place below after grist & prophecies

gabriel-v (changes requested):
asking for some small changes around configuration defaults and validity checks

templates/bbb.nomad
thifranc: I've left that one aside, because I think we're good with 2Gb of ram, the 8Go recommanded are for the whole BBB+greenlight server, as it's just the frontend here, I think we're OK with only 2Go of RAM, we can up to 4Go maybe, but I don't deem it necessary

PR #360, opened 2024-01-19
Office integration

k-jell:
closes https://github.com/CRJI/EIC/issues/1014


gabriel-v:
found a couple of problems with the collection creation UI

- message should also say min collection name length is 3
- the "already indexed" error happens when no nextcloud collections exist

Screenshot from 2024-03-12 15-57-12

templates/collabora.nomad
```diff
@@ -0,0 +1,85 @@
...
+
+job "collabora" {
+  datacenters = ["dc1"]
+  type = "service"
+  priority = 98
+
+  group "collabora" {
+    ${ group_disk() }
+    ${ continuous_reschedule() }
+    network {
```
gabriel-v, line 11: plz keep the member order in more or less the same order as the other templates - network goes at the bottom of the group, alongside other network-y stuff like resources and services.

Also, why is this under group > network and not task > resources > network  + task > configs > port_map? The more consistent we have these, the easier it will be to eventually migrate off this platform into another one.

templates/nextcloud28-deps.nomad
```diff
@@ -0,0 +1,165 @@
...
+        MINIO_API_REQUESTS_DEADLINE = "2m"
+        GOMAXPROCS = "20"
+      }
+
+      resources {
+        memory = 1500
+        cpu = 1000
+        network {
+          mbits = 1
+          port "s3" { static = 9005 }
```
gabriel-v, line 118: do we really need static ports here? they can be auto-assigned by nomad, which leaves us with less ports in the source code

templates/nextcloud28-deps.nomad
```diff
@@ -0,0 +1,165 @@
...
+          type = "tcp"
+          interval = "${check_interval}"
+          timeout = "${check_timeout}"
+        }
+      }
+
+      service {
+        name = "nextcloud28-minio-console"
+        port = "console"
+        tags = ["fabio-/nextcloud28_minio_console strip=/nextcloud28_minio_console"]
```
gabriel-v, line 153: do we really care about hosting the minio console UI on fabio?  have you successfully used it? I think the `/nextcloud28_minio_console` prefix we put on the URL will cause it to break (as will most web apps that you change the root url of)
\

templates/nextcloud28-entrypoint.sh
```diff
@@ -0,0 +1,125 @@
...
+],
+  'dbname' => getenv('MYSQL_DATABASE'),
+  'dbpassword' => getenv('MYSQL_PASSWORD'),
+);
+DELIM
+
+# Launch canonical entrypoint
+/entrypoint.sh apache2-foreground
+
+unzip /apps-to-install/sociallogin.zip -d /apps-to-install
```
gabriel-v, line 28: this should be moved in the docker build process, we don't want to be unzipping stuff every container start...

templates/nextcloud28-entrypoint.sh
```diff
@@ -0,0 +1,125 @@
...
+mkdir -p /var/www/html/custom_apps/sociallogin
+mv /apps-to-install/nextcloud-social-login-*/* /var/www/html/custom_apps/sociallogin
```
gabriel-v, line 30: check if already installed and don't install again, to avoid doing 200MB of I/O every container start -- better yet do this only once in initial installation step

templates/nextcloud28-entrypoint.sh
```diff
@@ -0,0 +1,125 @@
...
+rm -r /apps-to-install/nextcloud-social-login-*
+rm /apps-to-install/*.zip
```
gabriel-v, line 32: leave the zip/source files alone, we don't care about a little 200mb more than we care about not making any extra I/O on container start

templates/nextcloud28-entrypoint.sh
```diff
@@ -0,0 +1,125 @@
...
+"defaultGroup":"",
+"groupMapping":{"$OAUTH2_ADMIN_GROUP":"admin"}}
+]}'
+DELIM
+)
+
+run_as "php occ config:app:set sociallogin custom_providers --value=$OAUTH_SETTINGS"
+
+
+# run_as 'php occ config:app:set sociallogin auto_create_groups --value=1'
```
gabriel-v, line 75: explicitly set this to what you want it to be (1 or 0) to override any breakage users make by manually clicking on the admin button.

templates/nextcloud28-entrypoint.sh
```diff
@@ -0,0 +1,125 @@
...
+run_as 'php occ app:install contacts'
+run_as 'php occ app:install richdocumentscode'
+run_as 'php occ app:enable richdocumentscode'
+run_as 'php occ app:install richdocuments'
+run_as 'php occ app:enable richdocuments'
+
+run_as 'php occ config:app:set richdocuments disable_certificate_verification --value="yes"'
+# http://nextcloud28.joseph.alt-f4.ro/custom_apps/richdocumentscode/proxy.php?req=/hosting/discovery
+run_as 'php occ config:app:set richdocuments public_wopi_url --value=http://nextcloud28.liquid.example.org/custom_apps/richdocumentscode/proxy.php?req='
+run_as 'php occ config:app:set richdocuments wopi_url --value=http://nextcloud28.liquid.example.org/custom_apps/richdocumentscode/proxy.php?req='
```
gabriel-v, line 114: domain plz

gabriel-v (changes requested):

templates/nextcloud28-entrypoint.sh
```diff
@@ -0,0 +1,125 @@
...
+<?php
+$CONFIG = array (
+  'dbhost' =>  getenv('MYSQL_HOST'),
+  'trusted_domains' => [
+                    "localhost",
+                    getenv('NEXTCLOUD_HOST'),
+                    getenv('LIQUID_BASE_IP')
+],
+  'dbname' => getenv('MYSQL_DATABASE'),
+  'dbpassword' => getenv('MYSQL_PASSWORD'),
```
gabriel-v, line 21: another thing to add here is the website domain (that changes pretty regularly and causes the config.php bug)

templates/nextcloud28-entrypoint.sh
```diff
@@ -0,0 +1,125 @@
...
+run_as 'php occ config:app:set files_sharing outgoing_server2server_share_enabled --value="no"'
+
+run_as 'php occ config:system:set social_login_auto_redirect --value=true'
+run_as 'php occ config:system:set skeletondirectory'
+run_as "php occ config:app:set sociallogin custom_providers --value=$OAUTH_SETTINGS"
+
+run_as 'php occ app:install calendar'
+run_as 'php occ app:install polls'
+run_as 'php occ app:install contacts'
+run_as 'php occ app:install richdocumentscode'
```
gabriel-v, line 106: remove & uninstall onlyoffice and onlyoffice code stuff - we only want collabora

liquid_node/configuration.py
```diff
@@ -72,13 +71,15 @@ class Configuration:
...
         'matrix': 'is the new chat app',
+        'nextcloud28': 'has a file share system and a contact list of users.',
     }
 
     APP_REDIS_IDS = {
         'hoover': 1,
         'dokuwiki': 2,
         'codimd': 3,
         'nextcloud': 4,
+        'nextcloud28': 5,
```
gabriel-v, line 82: missing collabora proxyredis id = 7

gabriel-v:
cannot merge pls rebase

gabriel-v:
nvm did the rebase myself

https://github.com/liquidinvestigations/node/compare/office-integration-squish?expand=1

thanks!

PR #365, opened 2024-04-24
ignore collectiondocumenthit when backing up pg database

k-jell:
closes CRJI/EIC#1196

### DIFF
--- a/liquid_node/backup.py
+++ b/liquid_node/backup.py
@@ -21,15 +21,15 @@ SNOOP_ORIGINAL_DATA_ALLOC = "hoover-deps:minio-collections"
 HTTP_PORT = config.port_lb
 # list of postgres databases that should not be backed up.
 # the format should be <schema>.<table name>
-DATABASES_TO_IGNORE = [
+SNOOP_PG_TABLES_TO_IGNORE = [
     'public.common_data_collectiondocumenthit',
 ]
 
 
-def databases_to_ignore_str():
+def pg_tables_to_ignore_str(ignore_list):
     res = []
-    for db in DATABASES_TO_IGNORE:
-        res.append(f'-T {db}*')
+    for table in ignore_list:
+        res.append(f'-T {table}*')
     return ' '.join(res)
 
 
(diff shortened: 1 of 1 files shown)


liquid_node/backup.py
```diff
@@ -19,6 +19,18 @@
 SNOOP_BLOBS_DATA_ALLOC = "hoover-deps:minio-blobs"
 SNOOP_ORIGINAL_DATA_ALLOC = "hoover-deps:minio-collections"
 HTTP_PORT = config.port_lb
+# list of postgres databases that should not be backed up.
+# the format should be <schema>.<table name>
+DATABASES_TO_IGNORE = [
```
gabriel-v, line 24: this file works with 10 databases in 5 services, which one are we talking about?

plz rename const name to something like "SNOOP_PG_TABLES_TO_IGNORE" or something similarly specific

liquid_node/backup.py
```diff
@@ -228,9 +240,9 @@ def backup_pg(dest_file, username, dbname, alloc):
     tmp_file = Path(str(dest_file) + '.tmp')
     log.info(f"Dumping postgres from alloc {alloc} user {username} db {dbname} to {tmp_file}")
     cmd = (
-        f"set -exo pipefail; ./liquid dockerexec {alloc} "
-        f"pg_dump -U {username} {dbname} -Ox "
-        f"| gzip -1 > {tmp_file}"
+        f'set -exo pipefail; ./liquid dockerexec {alloc} '
+        f'pg_dump {databases_to_ignore_str()} -U {username} {dbname} -Ox '
```
gabriel-v, line 244: this applies this complication to all postgres databases.

we don't want to do that (even if we know for sure there'll be no table with the same name in the future) because it makes the whole process more convoluted and confusing to debug

plz only apply the ignore option to the actual database we want to work with (snoop)

gabriel-v (changes requested):

liquid_node/backup.py
gabriel-v: maybe some `ignore_tables = []` optional arg to the `backup_pg(...)` function

gabriel-v (approved):

